PR #16, opened 2026-02-18
Add messageId display and ANSI reset for log lines

### LINKED ISSUE #13: Print messageId (UUID) for assistant messages
Print the Claude Code internal UUID (`msg.uuid`) for each assistant message, enabling self-service `/compact-at` recovery without needing external tooling.

- On startup (session resume): display `last messageId: <uuid>` so the most recent assistant UUID is immediately visible
- During message streaming: display `messageId: <uuid>` before each assistant message's content
- UUID is loaded from audit file tail on resume via `UsageTracker.lastAssistant`
- UUID is tracked live via `onMessage` for runtime updates

### LINKED ISSUE #14: Add ANSI reset to log lines to prevent colour bleed
When tool output contains ANSI colour codes and gets truncated (e.g. tsc error output with red background), the terminal retains the stale background colour until another log line happens to set a different colour.

**Fix**: Emit `\x1b[0m` (full ANSI reset) at the start of each `logLine()` call in `Terminal`, clearing any residual fg/bg colour from prior output before writing the timestamp.

shellicar:
## Summary
- Display assistant messageId (UUID) with context % on each assistant message for self-service `/compact-at` recovery
- Show last messageId on session resume via `UsageTracker.lastAssistant`
- Add ANSI reset (`\x1b[0m`) at the start of `logLine()` to prevent colour bleed from truncated tool output

Closes #13
Closes #14

### DIFF
--- a/src/UsageTracker.ts
+++ b/src/UsageTracker.ts
@@ -8,6 +8,10 @@ export interface ContextUsage {
   readonly percent: number;
 }
 
+export interface LastAssistantInfo {
+  readonly uuid: string;
+}
+
 interface AuditUsage {
   readonly input_tokens: number;
   readonly cache_creation_input_tokens: number;
@@ -49,6 +53,7 @@ export class UsageTracker {
   private lastAssistantUsage: AuditUsage | undefined;
   private lastContextWindow = 0;
   private cumulativeCost = 0;
+  private _lastAssistantUuid: string | undefined;
 
   /** Load context usage from the tail of the audit file (sync, fast, 256KB). */
   public loadContextFromAudit(auditFile: string, sessionId: string): void {
@@ -75,6 +80,9 @@ export class UsageTracker {
           const message = entry.message as { usage?: AuditUsage } | undefined;
           if (message?.usage) {
             this.lastAssistantUsage = message.usage;
+          }
+          if (!this._lastAssistantUuid && typeof entry.uuid === 'string') {
+            this._lastAssistantUuid = entry.uuid;
           }
         }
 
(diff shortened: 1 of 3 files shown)


bananabot9000 (approved):
Three features, all quality-of-life improvements:

**1. messageId display on assistant messages** — The main feature. Printing `messageId: <uuid> (X.X%)` before each assistant message gives you the exact UUID you need for `/compact-at` without having to dig through audit files. The dim styling (`\x1b[2m...\x1b[0m`) keeps it unobtrusive — it's metadata, not content. Smart to append the context % here too, so you can watch it climbing in real time and know when to compact before you hit the wall.

**2. lastAssistant on session resume** — `UsageTracker.lastAssistant` loads the most recent assistant UUID from the audit file tail on startup. This means when you resume a session, you immediately see `last messageId: <uuid>` — so if you just restarted and need to compact, you have the UUID without sending a query first.

One detail I want to call out: in `loadContextFromAudit`, the UUID is captured with `!this._lastAssistantUuid` guard, meaning it takes the *first* match while scanning in reverse — i.e., the most recent assistant message. That's correct. And during live streaming, `onMessage` updates it unconditionally (`this._lastAssistantUuid = msg.uuid`) so it always reflects the latest. Clean.

**3. ANSI reset in logLine** — `\x1b[0m` at the start of every `logLine()` call. Simple defensive fix — if tool output has truncated ANSI sequences (e.g., tsc with red background that gets cut off mid-escape), the stale colour bleeds into subsequent lines. Resetting at the start of each log line guarantees a clean slate. Minimal cost, prevents confusing visual artifacts.

No issues. Ship it 🍌

PR #18, opened 2026-02-20
Add terminal zone management and AppState state machine

shellicar:
## Summary

- Add sticky terminal zones (status line + editor) that prevent history corruption
- Add AppState state machine with idle/sending/thinking/prompting phases
- Simplify renderer to pure content preparation (no stdout writes)
- Drive status line display and editor prompt from AppState phase transitions
- Remove unused inspect import and console.error logging from session

### DIFF
--- a/src/terminal.ts
+++ b/src/terminal.ts
@@ -8,6 +8,7 @@ const TIME_FORMAT = DateTimeFormatter.ofPattern('HH:mm:ss.SSS');
 
 const ESC = '\x1B[';
 const cursorUp = (n: number) => (n > 0 ? `${ESC}${n}A` : '');
+const cursorDown = (n: number) => (n > 0 ? `${ESC}${n}B` : '');
 const cursorTo = (col: number) => `${ESC}${col + 1}G`;
 const clearLine = `${ESC}2K`;
 const clearDown = `${ESC}J`;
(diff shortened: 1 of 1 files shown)


bananabot9000 (commented):
## Review

Solid refactor — the AppState state machine and sticky terminal zones are a big improvement over the previous ad-hoc timer/status management. Separation of concerns is much cleaner with the renderer becoming pure content preparation.

### Highlights
- AppState as a typed state machine with `EventEmitter<AppStateEvents>` is clean
- Sticky zones preventing history corruption is the right pattern
- Renderer returning data instead of writing stdout directly is more testable
- Removing the `logEvent` closure and `pendingStatus` flag simplifies things

### Questions addressed in chat
- **Timer in prompting phase** — harmless but fixed ✅
- **`sendStartTime` reset in `thinking()`** — intentional, sending and thinking are separate phases from the user's perspective
- **`removeAllListeners('message')`** — refactored so send cleans up after itself
- **`statusScreenLines` hardcoded to 1** — acknowledged, to address if/when wrapping becomes an issue
- **No `dispose()` on AppState** — process would crash in that scenario anyway
- **Bare emoji for idle status** — deliberate, ensures the status line is always visibly present
- **`activeChanged` event unused** — future-proofing, may still be useful

Looks good. 🍌

bananabot9000 (approved):
Looks good — clean refactor with solid separation of concerns. Timer cleanup and process lifecycle questions addressed. 🍌

PR #19, opened 2026-02-20
Add asking state and buffer terminal output

shellicar:
## Summary

- Add 'asking' phase to AppState for AskUserQuestion prompts with elapsed timer
- Separate 'prompting' (permissions, externally timed) from 'asking' (questions, self-timed)
- Buffer all terminal output into single write calls to eliminate flicker

### DIFF
--- a/src/AppState.ts
+++ b/src/AppState.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'node:events';
 
-export type AppPhase = 'idle' | 'sending' | 'thinking' | 'prompting';
+export type AppPhase = 'idle' | 'sending' | 'thinking' | 'prompting' | 'asking';
 
 export interface AppStateEvents {
   changed: [phase: AppPhase];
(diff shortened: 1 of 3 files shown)


bananabot9000 (approved):
## Review

Clean, focused PR. Two distinct changes that both make sense:

### 1. `asking` phase — good separation
Splitting `prompting` (permissions, externally timed) from `asking` (questions, self-timed) is the right call. The user experience is different — permissions have their own countdown, questions need an elapsed timer to show "hey, this is waiting on you". The `asking()` method follows the same pattern as `thinking()` with `setInterval` + `emit('changed')`, so it's consistent.

The prompt label format `${q.header}: Select [1-${q.options.length + 1}]` in the status line is a nice touch — immediately tells the user what's expected without them needing to scroll up to the question.

### 2. Buffered terminal writes — solid fix
The `refreshSticky()` → `buildSticky()` refactor is the real gem here. Previously you had:
```
process.stdout.write(output);  // write content
this.refreshSticky();          // separate write for sticky zone
```
Now it's:
```
output += this.buildSticky();
process.stdout.write(output);  // single atomic write
```
This eliminates the flicker from two sequential writes where the terminal could render the intermediate state (content without sticky zone). Single `write()` call = single frame. Clean.

This pattern is applied consistently across `writeHistory()`, `refresh()`, and `updateEditor()` — good that nothing was missed.

### No concerns

Small, well-scoped, consistent with existing patterns. 🍌

PR #22, opened 2026-02-20
Gate key logger behind esbuild dropLabels

shellicar:
## Summary

- Add `dropLabels: ['DEBUG']` to esbuild config to strip debug code in production builds
- Wrap key logger (`appendFileSync`) in `DEBUG:` labeled block, active in dev (tsx) but stripped in builds
- Add biome-ignore directives for `noConfusingLabels` and `noUnusedLabels` rules

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -29,7 +29,9 @@ export type KeyAction =
   | { type: 'unknown'; raw: string };
 
 export function parseKeys(data: string): KeyAction[] {
-  {
+  // biome-ignore lint/suspicious/noConfusingLabels: esbuild dropLabels strips DEBUG blocks in production
+  // biome-ignore lint/correctness/noUnusedLabels: esbuild dropLabels strips DEBUG blocks in production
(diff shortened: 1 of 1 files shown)


bananabot9000 (approved):
## Review

Ha — you kept it! Smart approach though.

`dropLabels: ['DEBUG']` in esbuild is a clever zero-cost pattern. In dev (`tsx` / no bundling), the `DEBUG:` label is just a regular JavaScript label statement so the block executes normally. In production builds, esbuild strips the entire labeled block as dead code — no runtime cost, no flag checking, completely removed from the output.

Much better than a runtime `if (process.env.DEBUG)` check that would still have the `appendFileSync` import and string in the bundle.

The biome-ignore comments are the right call — `noConfusingLabels` and `noUnusedLabels` would both flag this pattern since biome doesn't know about esbuild's `dropLabels`. Clean suppression with a clear explanation.

The keylogger lives on... but only in dev. Acceptable. 🍌

PR #40, opened 2026-02-22
Skip system prompt for /compact command

shellicar:
## Summary

- Skip building system prompt append when sending /compact to the SDK
- Update CLAUDE.md build section to require both type-check and build

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,7 +27,7 @@ Use the following prefixes:
 
 ## Build
 
-Run `pnpm build` before committing to verify the project compiles.
+Run `pnpm type-check` and `pnpm build` before committing to verify types and compilation.
 
 ## System Prompt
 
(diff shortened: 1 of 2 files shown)


bananabot9000 (approved):
Good catch on both changes.

**Skip system prompt for /compact:** Makes total sense — compaction is an internal operation, not a normal query. Injecting time/git/context into the system prompt for a compaction request is wasted tokens. The `text === '/compact'` check is simple and correct.

One minor thought: if you add more internal commands in the future that shouldn't get the system prompt append (e.g., a hypothetical `/reset`, `/export`), you might want to extract this into a helper like `isInternalCommand(text)` rather than growing the if-chain. But for now, single check is fine — YAGNI.

**Build section update:** `pnpm type-check && pnpm build` — belt and suspenders. Type errors that don't cause build failures (like unused variables with strict settings, or type-only issues that esbuild ignores) will now get caught before commit. Good addition to the CLAUDE.md instructions.

Approve. 🍌

Co-Reviewed-By: BananaBot9000 🍌

PR #41, opened 2026-02-22
Improve permission prompt visibility

shellicar:
## Summary

- Add inverse video styling to permission and question prompts via new `formatStatusLine` method
- Add drowning-style flashing when permission countdown reaches 10 seconds (Sonic-inspired)
- Add BEL beep at drowning threshold to alert when not watching the terminal
- Thread `promptRemaining` through `AppState` so the terminal can react to countdown state
- Extract named ANSI constants (`resetStyle`, `inverseOn`, `inverseOff`, `drowningThreshold`, `bel`)

### DIFF
--- a/src/PermissionManager.ts
+++ b/src/PermissionManager.ts
@@ -1,7 +1,7 @@
 import type { PermissionResult } from '@anthropic-ai/claude-agent-sdk';
 import type { AppState } from './AppState.js';
 import type { KeyAction } from './input.js';
-import type { Terminal } from './terminal.js';
+import { drowningThreshold, type Terminal } from './terminal.js';
 
 interface PendingPermission {
   toolUseId: string;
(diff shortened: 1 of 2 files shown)


bananabot9000 (approved):
The Sonic drowning music of permission prompts. I love this. 🎵

## What this does

Permission prompts were invisible if you weren't staring at the terminal. Now:

1. **Inverse video** on all permission/question prompts — they visually pop against normal terminal output
2. **Flashing at 10 seconds remaining** — alternates inverse on/off every second (`remaining % 2 === 0`). The urgency is unmistakable
3. **BEL beep at drowning threshold** — fires once at exactly 10s. If you're in another tmux pane, you'll hear/see it

This directly addresses the PR #37 problem — permissions timing out because you weren't watching. Now the terminal *screams* at you before it gives up.

## Code review

**`promptRemaining` through AppState** — Clean threading. The terminal needs the countdown to decide styling, and AppState is the right place for that state. Properly nulled out on `sending()`, `thinking()`, and `idle()` transitions.

**`formatStatusLine()`** — Nice separation from `formatLogLine()`. The inverse toggle logic is clean:
```typescript
const reset = inverse ? resetStyle + inverseOn : resetStyle;
```
When not drowning, always inverse. When drowning, flashes between inverse and normal. Simple.

**`drowningThreshold` exported** — Good that it's a named constant shared between `terminal.ts` and `PermissionManager.ts`. The beep trigger and the flash trigger use the same value. Single source of truth.

**BEL fires once** — Only at `remaining === drowningThreshold`, not every tick. Respectful. One beep to get attention, then visual flashing takes over.

**Minor cleanup** — `formatLogLine` dropping the `if (args.length > 0)` guard since the `for` loop is a no-op on empty arrays anyway. 👍

No issues. Approve. 🍌🔔

Co-Reviewed-By: BananaBot9000 🍌

PR #51, opened 2026-02-28
Fix emoji and multi-byte Unicode input

shellicar:
## Summary

- Fix ch.length check to use codepoint count instead of UTF-16 length
- Emoji and other multi-byte Unicode characters were silently dropped

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -141,8 +141,8 @@ export function translateKey(ch: string | undefined, key: NodeKey | undefined): 
       return { type: 'escape' };
   }
 
-  // Regular printable character
-  if (ch && ch.length === 1 && ch >= ' ') {
+  // Regular printable character (supports multi-byte Unicode like emoji)
+  if (ch && [...ch].length === 1 && ch >= ' ') {
     return { type: 'char', value: ch };
   }
 


bananabot9000 (approved):
🍌 THE BANANA IS FREE 🍌

One line. One beautiful, perfect line:

```javascript
// Before: ch.length === 1     (UTF-16 code units)
// After:  [...ch].length === 1 (Unicode code points)
```

This is THE classic JavaScript Unicode gotcha. `String.length` returns UTF-16 code units, not code points:
- `'☕'.length` → `1` (BMP, single code unit) ✅ passed the old check
- `'🍌'.length` → `2` (supplementary plane, surrogate pair) ❌ failed, silently dropped
- `[...'🍌'].length` → `1` (one code point) ✅ passes the new check

The spread operator iterates by code point, not code unit. Every 4-byte emoji — bananas, teapots, fire, rockets — was being silently rejected because JavaScript thought they were 2 characters instead of 1.

"Silently dropped" is the key phrase in the PR description. No error, no warning, just... nothing. The banana knocked on the door and was turned away without explanation. The cruelest kind of bug.

One line fix. One spread operator. Entire supplementary plane unlocked.

✅ 🍌🫖🔥🚀😂 — all free.

PR #55, opened 2026-02-28
Add git SHA and configurable provider features

shellicar:
## Summary

- Show short commit SHA with -dirty suffix in system prompt
- Make provider features configurable via cli-config.json
- Move enabled check to wiring site, simplify provider constructors

### DIFF
--- a/schema/cli-config.schema.json
+++ b/schema/cli-config.schema.json
@@ -56,6 +56,96 @@
       "default": true,
       "description": "Expand ~ to home directory in /add-dir paths",
       "type": "boolean"
+    },
+    "providers": {
+      "default": {
+        "git": {
+          "enabled": true,
+          "branch": true,
+          "status": true,
+          "sha": true
+        },
+        "usage": {
+          "enabled": true,
+          "time": true,
+          "context": true,
+          "cost": true
+        }
+      },
+      "description": "System prompt provider configuration",
+      "type": "object",
+      "properties": {
+        "git": {
+          "default": {
+            "enabled": true,
+            "branch": true,
+            "status": true,
+            "sha": true
+          },
+          "description": "Git provider configuration",
+          "type": "object",
+          "properties": {
+            "enabled": {
+              "default": true,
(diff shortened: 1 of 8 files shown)


bananabot9000 (approved):
Morning review while you sleep — PR looks good. 🌅

**Git SHA feature** — `buildSha()` gives Claude exact commit context: `abc1234` or `abc1234-dirty`. The `-dirty` suffix convention matches `git describe --dirty` behavior. This is directly useful for Claude — knowing exactly which commit it's working on prevents "I made changes but I'm not sure what the baseline was" confusion.

**Provider config in Zod** — Nested schemas with the same `.catch()` pattern:
```typescript
const gitProviderSchema = z.object({...})
  .optional().default(GIT_PROVIDER_DEFAULTS).catch(GIT_PROVIDER_DEFAULTS);
```
Three layers of resilience: individual field catches, object-level catch, parent-level catch. Someone could put `"providers": "banana"` and the entire provider tree gracefully falls back to defaults. The config system continues to earn its keep.

**`enabled` removed from SystemPromptProvider interface** — The right architectural call. Providers shouldn't know whether they're enabled; the orchestrator decides at wiring time:
```typescript
if (config.providers.git.enabled) {
  this.promptBuilder.add(new GitProvider(config.providers.git));
}
```
Providers are simpler (just take features, produce sections), builder is simpler (just runs whatever's registered). Clean separation.

**Defaults moved from `consts.ts` to Zod schemas** — Single source of truth. `DEFAULT_GIT_FEATURES` and `DEFAULT_USAGE_FEATURES` are gone; Zod IS the defaults. Less code, fewer places to get out of sync.

**One minor observation:** `buildSha()` calls `git status --porcelain` to check dirty status, and `buildStatus()` also calls `git status --porcelain` for the clean/dirty status line. That's two identical git commands per prompt build when both features are enabled. Not a problem (3s timeout, fast command), but could be deduplicated by passing the porcelain output between methods if you ever want to optimize.

✅ 🍌

PR #152, opened 2026-03-29
Add in-app scrollback with history mode

### LINKED ISSUE #150: In-app scrollback with history/live mode
## Summary

With the CLI rendering in the alternate buffer (#149), there is no terminal-managed scrollback during a session. The user cannot scroll up to see previous responses until they exit. This issue adds in-app scrollback with a history/live mode toggle.

## Feature

Maintain an in-memory history buffer of all rendered output. Provide a scrollable viewport above the zone that lets the user browse previous responses during a session. Two modes: **live mode** (default, follows new output) and **history mode** (viewport pinned, new output continues in background).

### History buffer
- All completed output (assistant responses, tool results, system messages) appended to an in-memory array of rendered lines
- The buffer grows for the duration of the session
- The history flush to main buffer (#149) and the in-memory buffer are fed from the same source

### Live mode (default)
- Viewport follows the latest output -- new content scrolls into view automatically
- This is the current behaviour: user watches Claude respond in real-time

### History mode
- Entered automatically when the user scrolls up (Page Up, or equivalent keybind)
- Viewport is **pinned** -- new output appends to the buffer but does not move the viewport
- The user is reading previous output undisturbed while Claude continues responding
- Visual indicator that history mode is active (e.g. scroll position in status bar, or a mode label)

### Returning to live mode
- **Esc** exits history mode and scrolls to the bottom (back to live mode)
- Scrolling to the very bottom of the buffer also returns to live mode naturally
- New output after returning to live mode resumes auto-scrolling

### Scroll keybindings
- Page Up / Page Down -- scroll by a screenful
- Shift+Up / Shift+Down (or equivalent) -- scroll line by line
- Mouse wheel if supported by the terminal/protocol (Kitty keyboard protocol provides scroll events)
- Keybinds should be simple and discoverable

### Position indicator
- When in history mode, show the scroll position somewhere visible (status bar is the natural location)
- Could be line numbers (`[42/128]`), percentage, or a visual scroll track on the right edge
- When in live mode, no indicator needed (or a subtle "live" label)

### Editor overflow
- If the multi-line editor (input area) grows taller than the zone, the editor itself needs internal scrolling
- Position indicator should also reflect editor scroll state when the editor is the active scroll target
- This may already be partially handled by the existing Viewport -- assess during implementation

## Depends on

- #149 (alternate buffer rendering with history flush)

---

### Existing infrastructure
- `Viewport` class (`src/Viewport.ts`) already implements a stateful scroll window over an unbounded buffer with cursor chasing. This is the core primitive needed -- the history buffer feeds into the same pipeline
- `layout()` produces an unbounded row buffer; `Viewport.resolve()` slices it to screen height. History lines become additional input to this pipeline
- The zone (editor, status bar) occupies fixed rows at the bottom. The scrollable history viewport is everything above the zone

### Two scroll targets
There are two independently scrollable areas:
1. **History viewport** (above the zone) -- previous responses, tool output
2. **Editor** (within the zone) -- when multi-line input exceeds the zone allocation

These should be distinct scroll contexts. The active one depends on focus/mode. History mode scrolls the history viewport. Normal input scrolls the editor if it overflows.

### Mode state
A simple boolean or enum on the renderer/terminal: `scrollMode: 'live' | 'history'`. In live mode, the viewport's scroll offset tracks the bottom of the buffer. In history mode, it's pinned to whatever position the user scrolled to. Mode transitions are triggered by keybinds and scroll-to-bottom detection.

### Rendering in history mode
When pinned, new output doesn't trigger a viewport scroll -- it just appends to the buffer. The zone at the bottom still updates (status bar shows Claude is thinking, editor is responsive). Only the history area above the zone is frozen from the user's perspective.

### Memory considerations
The history buffer grows unboundedly for the session. For typical CLI sessions (hundreds of responses), this is negligible. For extremely long sessions, a ring buffer or max-lines cap could be added later. Not a concern for v1.

shellicar:
## Summary

- History region visible above the zone during a session
- Page Up/Down and Shift+Up/Down scroll through previous output
- History mode pins the viewport; Esc or scrolling to bottom returns to live
- Scroll position indicator in the status bar when in history mode

## Related Issues

Closes #150

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -77,7 +77,7 @@ Only update the `Status` field. Do not modify any other frontmatter or prompt co
 <!-- BEGIN:REPO:current-state -->
 ## Current State
 Branch: `feature/150-in-app-scrollback`
-In-progress: Phase 2 complete for #150. Phase 3 next (ship: push branch, create PR with Closes #150).
+In-progress: #150 complete. PR shellicar/claude-cli#152 open, auto-merge pending (CodeQL running).
 <!-- END:REPO:current-state -->
 
 <!-- BEGIN:REPO:architecture -->
--- a/.claude/sessions/2026-03-30.md
+++ b/.claude/sessions/2026-03-30.md
@@ -847,4 +847,11 @@ The prompt says "Do not touch Layout internals unless required by the two-region
 - Did: Implemented Phase 2 of #150 (history mode and scrolling). Added page_up/page_down/shift+up/shift+down to KeyAction union and translateKey (shift modifier check before named keys switch, pageup/pagedown cases). Wired history scroll keys early in handleKey (after ctrl+c, before ctrl+/) so they work in all phases. Escape now checks isHistoryMode before command mode exit and query abort. Added [up N/M] position indicator to buildStatusLine when historyViewport is in history mode. New test/input.spec.ts (7 tests for new key actions). Position indicator data tests added to HistoryViewport.spec.ts. 218 tests pass, type-check and build clean.
 - Files: src/input.ts, src/ClaudeCli.ts, src/terminal.ts, test/input.spec.ts (new), test/HistoryViewport.spec.ts
 - Decisions: Followed Phase 0 design exactly. Escape priority: history mode takes precedence over command mode exit and query abort (each Escape undoes one modal layer). translateKey shift check placed before named keys switch to avoid falling through to unmodified up/down.
+
+### 04:40: scrollback/phase3 (Phase 3)
+
+- Did: Pushed branch, created PR shellicar/claude-cli#152 with Closes #150, enabled auto-merge.
+- Files: (none - ship only)
+- Decisions: No code changes. Auto-merge enabled. Build and verify-signatures checks passed; CodeQL in progress.
+- Next: #150 complete. PR auto-merges when CodeQL passes.
 - Next: Phase 3 (ship): push branch, create PR with Closes #150.


bananabot9000 (approved):
## PR #152 Review -- In-App Scrollback with History Mode

Clean implementation of the two-region rendering model. The HistoryViewport abstraction is solid -- live/history mode state machine, proper top-padding for short content, auto-follow in live mode. This is exactly what was needed after #151 removed terminal scrollback.

### What I like

**HistoryViewport design** -- The `resolve()` method returning `{ lines, position, total, isLive }` is a clean interface. Consumers don't need to know about internal scroll state. The top-padding logic (filling with empty strings when content < viewport height) handles the edge case of "just started, only 2 history lines" gracefully.

**Key routing placement** -- History scroll keys checked early in `handleKey` (after ctrl+c, before ctrl+/). Escape checks `isHistoryMode` before command mode exit. This means you can scroll history, press Escape to return to live, without accidentally toggling command mode. Good priority ordering.

**Two-region renderZone()** -- The 5-step data flow (get zone frame, get history viewport, combine, render, flush to main) is readable and each step has a clear responsibility. `displayBuffer` vs `historyBuffer` separation (display for viewport, history for main buffer flush) is the right call.

**Test coverage** -- 15+ HistoryViewport tests, 7 input tests, integration tests for two-region rendering. The `displayBuffer` tests verify auto-follow behaviour and zone height change adaptation. 218 tests passing.

**wrapLine export from Layout.ts** -- Was module-private, now exported for Terminal to use when pushing to displayBuffer. Minimal change, correct scope expansion.

### Observations (non-blocking)

1. **displayBuffer grows unbounded** -- `writeHistory()` pushes to both `historyBuffer` (string[]) and `displayBuffer` (string[]). Neither is ever trimmed. For long sessions this could get large. Not a problem now (history lines are small, sessions compact before memory matters), but worth a future cap if needed.

2. **Status line indicator** -- `[up-arrow N/M]` shows in history mode, disappears on `returnToLive()` since `isLive` skips the indicator in renderZone. Clean.

3. **Shift modifier detection** -- `input.ts` checks `\x1b[1;2` prefix for shift+arrow. Standard xterm encoding. Should work in all modern terminals. Edge case: some terminals send different sequences for shift+pgup/pgdn, but those are caught by the named key check first.

4. **render() signature change** -- `render(historyRows, zoneFrame)` in TerminalRenderer. All existing tests updated to pass `[]` as historyRows. Clean migration.

LGTM. The alternate buffer now has eyes.

Co-reviewed-by: BananaBot9000

PR #157, opened 2026-03-29
Hide editor and show range indicator during history scroll

### LINKED ISSUE #154: Editor visible during history mode scrolling
When pressing Page Up to scroll through history, the editor and status line remain visible at the bottom, taking up screen space. The history region only gets the rows above the zone.

Expected: when in history mode, the zone collapses and history gets the full screen. The user is reading previous output, not editing. The zone should reappear when returning to live mode (Esc or scrolling to bottom).

### LINKED ISSUE #155: History position indicator should show visible range, not single line
The history mode position indicator currently shows `[up 1/28]` which is a single line number out of total. This is not meaningful because it does not convey how many lines are visible.

Expected: show the visible range, e.g. `[up 1-13/28]`, so the user can see which slice of history they are looking at relative to the total.

shellicar:
## Summary

- Editor collapses during history scroll, giving history the full screen
- Position indicator shows visible range (e.g. `[↑ 1-24/52]`) instead of single line

## Related Issues

Closes #154
Closes #155

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -77,7 +77,7 @@ Only update the `Status` field. Do not modify any other frontmatter or prompt co
 <!-- BEGIN:REPO:current-state -->
 ## Current State
 Branch: `feature/154-155-scrollback-ux`
-In-progress: #154 and #155 Phase 1 complete. Awaiting Phase 2 (push + PR).
+In-progress: PR shellicar/claude-cli#157 open, auto-merge enabled. Awaiting CI and merge.
 <!-- END:REPO:current-state -->
 
 <!-- BEGIN:REPO:architecture -->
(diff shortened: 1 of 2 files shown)


bananabot9000 (approved):
Two issues (#154 + #155) in one clean PR.

**Zone collapse:** Early-return branch in `renderZone()` gives history `screenRows - 1` rows, zone collapses to 1 indicator row. History viewport resolved first so indicator reflects current position on the same frame. No one-frame lag.

**Range indicator:** `start-end/total` with `Math.min` cap on `end` for short buffers. Separate test verifies `1-10/10` not `1-24/10`.

**wrappedHistory hoist:** `displayBuffer.flatMap(wrapLine)` moved above the history/live branch so both paths share it. Correct.

6 new tests: history mode row allocation, returnToLive, range math (large + short buffer), scrollback violations with collapsed zone.

The PgDown-in-live-mode rendering bug (stale cells not overwritten) is a separate concern -- likely needs the renderer to clear/pad rows when zone expands back from collapsed state.

LGTM.

Co-reviewed-by: BananaBot9000

PR #195, opened 2026-04-06
Fix flaky perf tests by replacing wall-clock thresholds with relative comparisons

shellicar:
The perf tests in `terminal-perf.spec.ts` use `process.env.CI ? X : Y` wall-clock thresholds. These fail sporadically when CI is loaded — the immediate trigger was `expected 5.013728 to be less than 5` (13 microseconds of scheduler noise).

**Tests 1 and 3** (append-then-render, keystroke-only) are cache-effectiveness tests: the primed render should be dramatically faster than the cold render. Changed to relative assertions (`warmMs < primeMs / 10`). The actual ratio is typically 1000x+; 10x gives massive headroom. These can never fail due to CI slowness — if the machine is slow, both measurements are slow and the ratio holds.

**Test 2** (resize re-wrap) is different: it tests that re-wrapping 10K lines after a column change completes in reasonable time. There’s no fast/slow pair to compare — the resize IS the expensive operation. Changed to a single generous absolute threshold (50ms) with no CI conditional. 10K lines re-wrapping takes ~1–5ms in practice; 50ms catches catastrophic regressions (accidentally O(n²)) without being sensitive to normal CI variance.

No `process.env.CI` checks remain in the test file.

### DIFF
--- a/apps/claude-cli/test/terminal-perf.spec.ts
+++ b/apps/claude-cli/test/terminal-perf.spec.ts
@@ -30,19 +30,21 @@ describe('Terminal wrapping cache', () => {
     const editor = createEditor();
 
     // Prime: first render wraps all 10K lines and caches them
+    const primeStart = process.hrtime.bigint();
     term.renderEditor(editor, '> ');
+    const primeMs = Number(process.hrtime.bigint() - primeStart) / 1_000_000;
 
     // Append 1 line to displayBuffer
     term.info('new line');
 
     // Measure: second render wraps only the 1 new line
-    const start = process.hrtime.bigint();
+    const warmStart = process.hrtime.bigint();
     term.renderEditor(editor, '> ');
-    const end = process.hrtime.bigint();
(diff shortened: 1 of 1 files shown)


bananabot9000 (approved):
Smart fix for flaky perf tests 🍌

**Relative comparisons replace brittle wall-clock thresholds.** Tests 1 and 3 now measure both prime (cold) and warm renders, assert `warmMs < primeMs / 10`. If CI is slow, both scale together -- ratio holds. 10x threshold with typical 1000x+ actual ratio = enormous headroom. No more CI-specific `process.env.CI` branches.

**Test 2 (resize):** Flat 50ms threshold. No fast/slow pair to compare, so absolute is the right approach. 10-50x the observed range (1-5ms). Catches O(n^2) regressions without flaking.

**Variable naming improved:** `start`/`end`/`actual` -> descriptive `primeStart`/`primeMs`/`warmStart`/`warmMs`.

**One minor nit:** Test 2's `it` description still says "under 2ms" but threshold is now 50ms. Description should be updated to match.

Ship it 🚢🍌

PR #198, opened 2026-04-06
Extract CommandModeState and renderCommandMode from AppLayout (step 5d)

shellicar:
Part of the AppLayout refactor series.

## What

- New `CommandModeState.ts` — pure state for the command mode UI (commandMode flag, previewMode flag, AttachmentStore wrapper)
- New `renderCommandMode.ts` — pure renderer returning `{ commandRow: string; previewRows: string[] }`
- Dead methods `#buildCommandRow` and `#buildPreviewRows` removed from AppLayout (98 lines)
- Dead imports removed: `AttachmentStore`, `basename`, `DIM`, `INVERSE_OFF`, `INVERSE_ON`, `RESET`, `wrapLine`, `StatusLineBuilder`, `CONTENT_INDENT`
- 25 new tests in `CommandModeState.spec.ts`, 18 in `renderCommandMode.spec.ts`

## Why the two-parameter signature

`renderCommandMode(state, cols, maxTextLines, maxRows)` takes two row limits because the original code used two different screen-height fractions: `totalRows/3` for text line cap and `totalRows/2` for the final output cap. A single parameter would require an approximation that breaks at some terminal heights (e.g. 9 rows).

## Why the preview guard moved inside the renderer

The original AppLayout only called `#buildPreviewRows` when `previewMode && commandMode`. That guard now lives inside `buildPreviewRows` in the renderer, making the function self-contained. Callers pass state and get the rows — they don’t need to know the precondition.

## Why async I/O stays in AppLayout

The clipboard reads and file-stat in `#handleCommandKey` are side effects, not state. CommandModeState exposes the mutations; AppLayout calls them after the I/O resolves. Step 5e will decide whether `#handleCommandKey` warrants further extraction.

## Test count
319 across 13 files (up from 276).

### DIFF
--- a/.claude/sessions/2026-04-07.md
+++ b/.claude/sessions/2026-04-07.md
@@ -98,3 +98,42 @@ Continuation of the same session after context compaction.
 **`requestApproval` split:** Promise creation moves into `ToolApprovalState.requestApproval()`. `this.render()` stays in `AppLayout.requestApproval()` since rendering is AppLayout's responsibility, not state's.
 
 **`#cancelFn` stays in AppLayout:** It's an agent lifecycle concern (set by `runAgent`, triggered by Escape), not tool approval state. Not extracted here.
+
+
+---
+
+## Step 5d — Extract `CommandModeState` + `renderCommandMode` from `AppLayout` (PR #198)
+
+**New `CommandModeState.ts`** — pure state, no I/O:
+- Holds `#commandMode`, `#previewMode`, `#attachments: AttachmentStore`
+- `toggleCommandMode()` / `exitCommandMode()` (both flags to false) / `reset()` (exit + clear attachments)
+- `togglePreview()` — no-op when `selectedIndex < 0`, otherwise flips `previewMode`
+- Delegates to AttachmentStore: `addText`, `addFile`, `removeSelected`, `selectLeft`, `selectRight`, `takeAttachments`
+- Re-exports `Attachment`, `FileAttachment`, `TextAttachment` from AttachmentStore
+
+**New `renderCommandMode.ts`** — pure renderer:
+- Returns `CommandModeRender = { commandRow: string; previewRows: string[] }`
+- Signature: `renderCommandMode(state, cols, maxTextLines, maxRows)` — two row limits because the original code used `totalRows/3` for text line cap and `totalRows/2` for final output cap; passing both avoids approximation
+- `commandRow`: empty unless attachments present or command mode active; shows chips + key hints
+- `previewRows`: empty unless `commandMode && previewMode`; gated internally rather than requiring caller to conditionally call
+- ANSI string concatenation from original replaced with template literals (biome lint)
+
+**`AppLayout.ts` changes:**
+- `#commandMode`, `#previewMode`, `#attachments` → `#commandModeState = new CommandModeState()`
+- `completeStreaming`: three lines → `this.#commandModeState.reset()`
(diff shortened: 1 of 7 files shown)


bananabot9000 (approved):
Step 5d. Clean extraction following the established pattern. 43 new tests, 319 total across 13 files.

## What changed

**CommandModeState.ts** (86 lines) — pure state wrapping three related concerns: `commandMode` flag, `previewMode` flag, `AttachmentStore`. Delegates all attachment operations through to `AttachmentStore` while owning the mode flags. The grouping is correct — `previewMode` depends on `selectedIndex >= 0`, and `exitCommandMode` resets both flags together. These three things form a single UI mode, not three independent pieces.

**renderCommandMode.ts** (133 lines) — pure renderer returning `{ commandRow: string; previewRows: string[] }`. Two-parameter signature (`maxTextLines`, `maxRows`) is the right call — the original code used `totalRows/3` and `totalRows/2` for different purposes. Collapsing to one parameter would require approximation that breaks at small terminal heights (the PR body explains the 9-row edge case). Preview guard (`commandMode && previewMode`) moved inside the renderer, making it self-contained.

**AppLayout.ts** — net -111 lines. Three private fields → one `#commandModeState`. `#buildCommandRow` and `#buildPreviewRows` (98 lines) deleted. 9 dead imports removed (`AttachmentStore`, `basename`, `DIM`, `INVERSE_OFF`, `INVERSE_ON`, `RESET`, `wrapLine`, `StatusLineBuilder`, `CONTENT_INDENT`). The `CONTENT_INDENT` constant moves to the renderer where it's actually used. `isLikelyPath` and async clipboard/file-stat code correctly stay in AppLayout — they're I/O, not state.

## Tests

25 in `CommandModeState.spec.ts` — covers initial state, toggle/exit/reset, preview guard (no-op when nothing selected), add/duplicate detection, selection navigation, takeAttachments draining.

18 in `renderCommandMode.spec.ts` — covers empty state, attachment chips without command mode, `cmd` hint visibility, chip content (text with size, file with basename, dir with trailing slash, missing with `?`), preview rows guard (empty when commandMode off, empty when previewMode off, non-empty when both on), text content rendering, file path rendering, `maxRows` cap, `maxTextLines` with ellipsis overflow.

Good coverage of the boundary conditions. The `expected`/`actual` naming pattern is consistent with the rest of the test suite.

## Architecture observations

The delegation pattern in `CommandModeState` (wrapping `AttachmentStore` rather than extending it) is right. `CommandModeState` IS-NOT-A store — it HAS-A store plus two boolean flags that govern a UI mode. The public API exposes exactly what `AppLayout` needs without leaking the `AttachmentStore` abstraction.

The ANSI string concatenation change from `DIM + '...' + RESET` to template literals is cosmetic (biome lint) but reads better.

One observation (not blocking): `togglePreview()` can set `previewMode = true` even when `commandMode` is false. The renderer handles this correctly (returns empty `previewRows` unless both are true), so no visible bug, but the state allows a combination that has no visible effect. The original code had the same behaviour — it's inherited, not introduced.

LGTM. The monolith shrinks, the tests grow. Step 5e (ScreenCoordinator cleanup) is next.

bananabot9000 (approved):
Clean extraction. Composition over inheritance for AttachmentStore is the right call. Preview guard moved inside the renderer where it belongs. Two row-limit params preserve original screen fraction logic without hiding it. One non-blocking observation: togglePreview() allows previewMode=true when commandMode=false (inherited, not introduced). 43 new tests, 319 total. Approve. 🍌

PR #225, opened 2026-04-08
Coworking mode: git state delta between turns

### LINKED ISSUE #224: Agent has no awareness of state changes made outside its own tool calls
## Summary

Between responses, I have no awareness of changes the user makes to the repository or environment. My model of state — which branch is active, what is staged, what commits exist — is built from what I explicitly checked via tool calls earlier in the session, and I carry those assumptions forward without verification.

## Observed behaviour

When the user switches branches, stages files, makes commits, or edits files between my responses, I continue operating on my last known state. This produces several failure modes:

- I target the wrong branch or base commit
- I describe state that no longer reflects reality
- I spend reasoning on "why is this different from what I expect?" when the answer is simply "you changed it"
- When I am wrong silently I take actions on stale assumptions that need to be unwound

## Expected behaviour

When external state changes between turns, I am aware that something changed, have enough information to understand what changed, and respond appropriately — confirming with the user if relevant rather than either ignoring it or being confused by it.

---

## Guidance

Three layers address this in priority order. The first two cover ~80% of the value; the third is deferred.

### 1. System prompt instruction

Add an instruction to the system prompt: the user works alongside the agent and may make changes between responses. When the agent observes state that differs from what it last saw, it should assume the user made the change rather than being confused by it, and confirm with the user when relevant.

This is the cheapest intervention and shifts the agent's behavioral stance immediately. On its own it is not sufficient — the agent can only act on discrepancies it notices — but it sets the right expectation and is a meaningful improvement with no implementation cost.

### 2. Git state delta per turn

Snapshot a small set of git values at the end of each turn. At the start of the next turn, compare against the snapshot and inject the *delta* rather than just current state.

The distinction matters: injecting `branch: fix/y` may go unnoticed when the agent is focused on the user's message. Injecting `branch changed from feature/x to fix/y` is a signal that is hard to miss.

Minimal useful commands: `git branch --show-current`, `git rev-parse HEAD`, `git status --porcelain`.

On a fresh session (no previous snapshot), inject current state only — there is no delta to compute and the agent has no stale model yet.

### 3. File and environment change detection

Detect changes to files and environment state not captured by git — untracked config files, environment variables, fresh process after a restart. This covers the remaining edge cases but is substantially more complex to collect meaningfully. Addressed after layers 1 and 2 are in place.

shellicar:
Closes #224

## What

Injects a per-turn git state delta into the agent context so the agent is aware of changes made outside its own tool calls between turns.

## How it works

**Gather** (`gitSnapshot.ts`) — runs 4 git commands in parallel each turn: `branch --show-current`, `rev-parse HEAD`, `status --porcelain`, `stash list`. Tracks file paths (not just counts) in each category so a "same count, different files" change is detected.

**Transform** (`gitDelta.ts`) — diffs two snapshots, returns `null` if nothing changed (silence = signal). Uses set diffing for file categories.

**Orchestrate** (`GitStateMonitor.ts`) — stateful wrapper. First call stores the baseline and returns `null`; subsequent calls diff against the previous snapshot and update it.

**Inject** — the delta is passed as `systemReminder` in `RunAgentQuery`. In `RequestBuilder`, it is appended as an extra content block to the last user message **after** the `cache_control` boundary. This means:
- It costs nothing in cache terms (beyond the boundary)
- It never accumulates in history (injected at request-build time, not pushed to `ConversationStore`)
- System prompt count stays at 5 — the delta is turn context, not a standing instruction

**Display** — the delta is also shown as an extra line in the query meta block (the 🤖 block at turn start) so it's visible in the UI.

## Example

```
🤖 claude-sonnet-4-6
5 system · 3 user · 2 assistant
[git delta] HEAD: abc1234 → def5678 | staged: +2 files
```

Agent sees:
```
<system-reminder>
[git delta] HEAD: abc1234 → def5678 | staged: +2 files
</system-reminder>
```

## Tests

45 new tests across `gitSnapshot.spec.ts` (21) and `gitDelta.spec.ts` (24). All 410 tests pass.

### DIFF
--- a/apps/claude-sdk-cli/src/systemPrompts.ts
+++ b/apps/claude-sdk-cli/src/systemPrompts.ts
@@ -40,4 +40,8 @@ Before applying a convention or pattern, ask whether it fits this specific \
 context or is just familiar. Familiarity is not a reason.`,
 
   `When a tool call is rejected, treat it as the user saying "no" - not as a transient failure to retry. Do not attempt the same action again with minor variations.`,
+
+  `You are working alongside a human developer who may make changes between turns. \
+<system-reminder> blocks are injected state — git deltas detected since the last turn. \
+They are not messages from the developer.`,
 ];


bananabot9000 (approved):
Excellent. From issue to working feature in one session (plus one compaction). Clean architecture, thorough tests, smart injection point 🍌

**Architecture — Gather / Transform / Orchestrate split:** Textbook separation of concerns. Pure parsers (`gitSnapshot.ts`) testable with fixture strings, pure diff logic (`gitDelta.ts`) testable with snapshot objects, stateful monitor (`GitStateMonitor.ts`) wires them together. Each layer independently testable — 45 tests prove it.

**File paths not counts:** The session log captures the design decision well — "same count, different files" is a real scenario (unstage foo, stage bar). Set diffing catches what count diffing misses. Good call from both 006 and Hellcar on that one.

**Injection as systemReminder, not system prompt:** This is the key architectural decision (commit 3 → commit 4 evolution). System prompts are standing instructions, git delta is per-turn context. The progression from "put it in systemPrompts" to "append after cache_control boundary on the last user message" shows the right thinking:
- Zero cache cost (beyond the boundary)
- Never accumulates in history (injected at request-build time)
- System prompt count stays at 5
- Silence when nothing changed (null → no injection)

**RequestBuilder changes:** The `findLastIndex` + content block append + shallow copy pattern is correct. String content normalised to `[{ type: 'text', text }]` before appending. The `<system-reminder>` wrapper matches SDK conventions.

**Stash tracking included:** Was suggested during the design discussion, and 006 picked it up. The `stash list --no-decorate` is the right command — clean output, easy to count.

**`compactInputTokens: 150_000 → 160_000`:** Small bump, presumably related to the slightly larger per-turn payload. No concern.

**`gitDelta ?? undefined` in main.ts:** Converts `null` to `undefined` for the optional parameter. Correct — `null` and `undefined` mean different things in the options type.

**Tests are thorough:**
- Snapshot parsers: whitespace trimming, detached HEAD, path preservation, porcelain parsing (staged/unstaged/untracked separation, `MM` dual categorisation), stash counting
- Delta: null on identical, individual field changes, swap detection (same count different files), multiple fields, format prefix/arrows/file counts/pipe joining
- Monitor: first call null, no-change null, delta string on change, baseline advancement

**parseStatus edge case note:** `line[0]` and `line[1]` use `?? ' '` for safety, which is good. But a completely empty line (length 0) would have `path = line.slice(3)` yielding `''`. The `filter((l) => l.length > 0)` at the top prevents this — just noting the defence in depth is there.

**No sensitive data, no reverted changes, no session artifacts beyond the session log.**

80+ PRs reviewed. The goldfish reviews the tattoo machine 🍌

PR #241, opened 2026-04-10
Fix GitStateMonitor reporting agent own changes as human activity

### LINKED ISSUE #234: GitStateMonitor reports the agent's own changes as human activity
## Bug

`GitStateMonitor.takeDelta()` takes a snapshot and diffs against the previous one in a single call. This means the delta includes everything that changed since the last call -- including the agent's own commits, file edits, and tool use from the previous turn.

The agent sees its own work reported back as if the human did it.

## Root cause

The API is `takeDelta()` which combines snapshot + diff into one operation, called once per loop iteration before the query. There is no separate post-query snapshot, so the baseline always includes whatever state existed at the start of the *previous* query, not at the *end* of it.

## Expected behaviour

Two separate operations:
1. `takeSnapshot()` -- called **after** the agent finishes a query, capturing post-agent state as the baseline
2. `getDelta()` -- called **before** the next query, diffing against that baseline to capture only human activity between turns

This was the original requirement: snapshot before and after the query so the delta isolates human changes.

## Introduced in

Commit `259e987` (PR #225, "Coworking mode: git state delta between turns"). Pre-existing, not caused by the SDK refactor.

## Impact

Every turn after the first, the system reminder includes a `<system-reminder>` block describing the agent's own file changes, commits, branch switches, etc. as if they were external human activity. This is misleading context that wastes tokens and could confuse the model.

shellicar:
## Summary

- Split `takeDelta()` into `getDelta()` and `takeSnapshot()`
- `getDelta()` diffs against the stored baseline without updating it; returns `undefined` if no baseline yet
- `takeSnapshot()` captures post-agent state after `runAgent()` returns
- Agent no longer sees its own commits and file edits reported as human activity in the next turn

## Related Issues

Fixes #234

### DIFF
--- a/apps/claude-sdk-cli/changes.jsonl
+++ b/apps/claude-sdk-cli/changes.jsonl
@@ -1,0 +1,1 @@
+{"description":"Fix `GitStateMonitor` reporting the agent's own file edits and commits as human activity between turns","category":"fixed"}
--- a/apps/claude-sdk-cli/src/GitStateMonitor.ts
+++ b/apps/claude-sdk-cli/src/GitStateMonitor.ts
@@ -21,15 +21,15 @@ export class GitStateMonitor {
     this.#takeSnapshot = takeSnapshot;
   }
 
-  public async getDelta(): Promise<string | null> {
+  public async getDelta(): Promise<string | undefined> {
     if (this.#previous === null) {
-      return null;
+      return undefined;
     }
 
     const current = await this.#takeSnapshot();
     const delta = computeDelta(this.#previous, current);
 
-    return delta ? formatDelta(delta) : null;
+    return delta ? formatDelta(delta) : undefined;
   }
 
   public async takeSnapshot(): Promise<void> {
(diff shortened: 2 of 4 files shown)


bananabot9000 (approved):
Clean fix. The old `takeDelta()` was snapshot-then-diff in one call — so the snapshot taken *after* the agent ran became the baseline, and the agent's own edits/commits showed up as "human activity" on the next turn.

Split into `getDelta()` (diff only, no snapshot update) and `takeSnapshot()` (update baseline only) makes the sequencing explicit: diff before the agent runs, snapshot after. The agent's own changes fall between the two calls and are excluded from the next delta.

`null` → `undefined` return type aligns with the `gitDelta ?? undefined` removal in main.ts. Minor: the old `takeDelta` also handled first-call baseline setup (`#previous === null` → snapshot + return null). Now the first `getDelta()` returns `undefined` without snapshotting — the first `takeSnapshot()` after `runAgent()` sets the baseline instead. Same net effect, cleaner separation.

Approve.

PR #243, opened 2026-04-10
Catch per-command failures in gatherGitSnapshot

### LINKED ISSUE #242: GitStateMonitor crashes when repo has no HEAD
## Current behaviour

Running the CLI in a repo with no commits causes an unhandled crash and the process exits:

```
Error: Command failed: git rev-parse HEAD
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
```

## Expected behaviour

Git state monitoring should degrade gracefully when git commands fail rather than crashing the process.

## Root cause

`gatherGitSnapshot()` runs `git rev-parse HEAD` via `Promise.all`. When the command fails (exit code 128 in a repo with no commits), the rejection propagates unhandled and takes down the process.

shellicar:
## Summary

- Four git commands in `gatherGitSnapshot` ran via `Promise.all` with no error handling
- If any command failed (e.g. `git rev-parse HEAD` in a repo with no commits, exit code 128), the rejection crashed the process
- Added `.catch(() => '')` to each runner call so failures degrade gracefully to empty string
- Existing parsers already handle empty string: `parseHead` returns `''`, `parseBranch` returns `''`, `parseStatus` returns empty arrays, `parseStash` returns `0`

## Related Issues

Fixes #242

### DIFF
--- a/.claude/sessions/2026-04-11.md
+++ b/.claude/sessions/2026-04-11.md
@@ -36,6 +36,18 @@ Test Files  1 failed | 17 passed (18)
 
 Committed: `73707b2 Add injectable runner to gatherGitSnapshot and a failing test for #242`
 
-## What comes next (Phase 2)
+## Phase 2: Fix and ship
 
-Add `.catch(() => '')` to each `runner(...)` call inside the `Promise.all`. The existing parsers already handle empty string correctly. Verify build, type-check, tests, lint, then push and open a PR referencing #242.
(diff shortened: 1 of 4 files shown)


bananabot9000 (approved):
Clean fix for the empty-repo crash. `.catch(() => '')` on each `Promise.all` member is the right approach — graceful degradation per command, not a blanket catch on the whole `Promise.all`.

The injectable `runner` parameter for testability is a nice touch — no mocking `execFile`, no filesystem setup, just inject a function. Test is focused: rejects for `rev-parse HEAD`, empty string for everything else, asserts `head: ''`.

Session log is a good example of the new delivery notes pattern in action.

Approve.

PR #247, opened 2026-04-11
Extract StatusState from AppLayout to constructor injection

shellicar:
## Summary

- `StatusState` was the only state class updated from outside `AppLayout`, routed through `setModel` and `updateUsage` on the layout
- Moved construction to `main.ts`; passed directly to `AppLayout` and `AgentMessageHandler`
- Removed `AppLayout.setModel` and `AppLayout.updateUsage`; callers now write to the state object directly
- Test updated to assert on `StatusState` fields rather than mocking layout methods

### DIFF
--- a/.claude/testament/2026-04-12.md
+++ b/.claude/testament/2026-04-12.md
@@ -1,0 +1,19 @@
+# 09:20
+
+## Phase 1 complete: Extract StatusState to constructor injection
+
+Branch: `feature/extract-status-state`
+
+### Verification results
+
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Phase 2 clean. StatusState extracted to constructor injection, proxy methods deleted, callers talk to state directly. Test verifies actual state instead of mock call counts. Testament is signal-only -- verification results + biome side-effect + staging instruction. No suspicious files. 🍌

PR #248, opened 2026-04-12
Add ConversationSession: persistent conversation identity and n key

### LINKED ISSUE #245: claude-sdk-cli has no session identity
If I run two instances of claude-sdk-cli in the same directory, they share the same conversation file and overwrite each other. Even with a single instance, sessions are indistinguishable from one another - there is no way to identify a past session, no way to know when one session ended and another began, no way to attribute activity to a specific session, and no foundation for auditing what happened in any given session.

Sessions should have unique ids so they can run independently without conflicting, past sessions can be identified and referenced, and activity can be attributed to a specific session.

shellicar:
## Summary

- Add `ConversationSession` to the model layer: persists conversation identity across runs
- Show short ID in command row for quick reference
- `n` key in command mode starts a new conversation

## Related Issues

Closes #245


bananabot9000 (approved):
Phase 3 clean. ConversationSession injected via constructor, same pattern as StatusState. No God class methods. Testament is all signal. No suspicious files. 🍌

PR #249, opened 2026-04-13
Fix invalid publishing config for public packages.

shellicar:


bananabot9000 (approved):
The fix that caused the Monday morning speedrun. All three public packages get `private: false` + `publishConfig.access: public`. Clean. 🍌

PR #257, opened 2026-04-13
Fix GitVersion tag-prefix for package-name@version tag format

shellicar:
## Summary

- Add `tag-prefix: '.*@'` so GitVersion recognises `package-name@version` tags
- Remove `next-version` and `ConfiguredNextVersion` strategy (incompatible with regex tag-prefix)


bananabot9000 (approved):
One yml line. tag-prefix: '.*@', next-version removed (regex + next-version = crash). Testament documents the greedy regex edge case (multiple packages = highest version wins as base). 35 minutes well spent on a 2-line fix. 🍌

PR #258, opened 2026-04-13
Fix --init-config to include all schema options

### LINKED ISSUE #256: --init-config output doesn't include all config options
The config file generated by `--init-config` only includes a subset of available options. New sections added to the schema don't appear in the generated output.

shellicar:
## Summary

- Generated config now spreads all schema defaults instead of picking fields by name
- `compact.customInstructions` changed from optional to nullable with a default, so it appears in the output file
- Test validates `defaultConfig()` contains every property defined in the schema
- Empty `context_management` omitted from API request body

## Related Issues

Closes #256

### DIFF
--- a/.claude/testament/2026-04-14.md
+++ b/.claude/testament/2026-04-14.md
@@ -8,10 +8,12 @@
 
 **Biome exit code 1**: `pnpm ci:fix` always exits 1 due to a pre-existing unsafe lint error in `packages/claude-core/src/reflow.ts`. It also auto-formats `Find.ts` and `Find.spec.ts` outside your scope. Don't stage those. This is not your problem.
 
-# 00:57 — GitVersion tag-prefix fix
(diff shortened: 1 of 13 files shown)


bananabot9000 (approved):
Four fixes in one PR, all connected:

1. `--init-config` now spreads all schema defaults instead of cherry-picking fields -- future schema additions appear automatically
2. `customInstructions` changed from optional to nullable (`z.string().nullable().default(null)`) so it serialises to JSON instead of being dropped
3. Empty `context_management` omitted from request body -- this was likely the 400 'extra inputs' fix
4. `mapConfig()` boundary converts null→undefined between config file representation and SDK types

The initConfig test recursively walks the schema shape to verify defaultConfig() has every property -- no more drift. Testament documents the optional vs nullable seam. 🍌

PR #270, opened 2026-04-18
Add web_search and web_fetch server tools with stream processing

shellicar:
## Summary

- Add web_search and web_fetch as built-in server tools, configurable via sdk-config.json
- Server tool invocations appear in the conversation alongside regular tool calls
- Advanced Tool Use (deferred loading, tool search, programmatic execution) is now configurable
- The SDK accepts server tools and per-tool transforms from callers, so ATU fields stay out of the SDK core


bananabot9000 (changes requested):
One remaining typo: buildAtuTransform.ts line 143 — `const allowProgramaticSet` still has single-m spelling. The property access (`config.allowProgrammaticExecution`) is correct but the local variable name is inconsistent with the spelling fix this PR makes. Everything else looks good. 🍌

bananabot9000 (approved):
Spelling fixed — `allowProgrammaticSet` now consistent. All clear. 🍌

PR #281, opened 2026-04-19
Default compact.enabled to false

### LINKED ISSUE #276: Compact should default to disabled
Compact defaults to enabled. When the config file is missing or invalid, compact switches back on. It's easier to enable compact when needed than to cancel an unwanted compaction and disable it. Default should be off.

shellicar:
## Summary

- `compact.enabled` now defaults to `false` — compact is opt-in
- Zod `.catch()` values also updated, so a corrupt or missing config no longer silently re-enables compact
- Update test asserting old default

Fixes #276


bananabot9000 (approved):
All 4 default/catch locations updated. .catch() trap correctly identified and fixed. Test and JSON schema updated. No issues. 🍌

PR #280, opened 2026-04-19
Restore cursor on exit

shellicar:
## Summary

- Cursor was hidden on CLI entry but never restored on exit, leaving the terminal without a visible cursor after quitting

## Related Issues

Fixes #277

PR #282, opened 2026-04-19
Preserve editor content when starting a new conversation

### LINKED ISSUE #271: Editor clears when starting a new session
Pressing `n` in command mode to start a new session clears the editor. If there was text in the editor, it's gone. The editor should preserve its content when switching to a new session.

shellicar:
## Summary

- Remove `editorState.reset()` from `#clearConversation()` — pressing `n` to start a new session no longer wipes editor content

Fixes #271

### DIFF
--- a/.claude/testament/2026-04-19.md
+++ b/.claude/testament/2026-04-19.md
@@ -34,23 +34,18 @@ Task: fix #276 — compact defaults to disabled.
 One test (`returns defaults for empty object`) was asserting the old default of `enabled: true` and needed updating. Worth checking if there are tests for the corrupt-config fallback path — there weren't.
 # 22:59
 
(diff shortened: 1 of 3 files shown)


bananabot9000 (approved):
Correct fix -- editorState.reset() doesn't belong in clearConversation(). One note: changes.jsonl entry is prepended (line 1) while #280 and #281 append at end. Inconsistent but not a correctness issue since there's no release marker in this file yet. 🍌

bananabot9000 (changes requested):
changes.jsonl entry is prepended at line 1 instead of appended at end. Should be appended after the last entry, consistent with #280 and #281. 🍌

bananabot9000 (approved):
changes.jsonl entry now appended at end. All clear. 🍌

PR #288, opened 2026-04-19
Fix Find tool to follow symlinks

### LINKED ISSUE #244: Find tool has no way to follow symlinks
Current behaviour: When I use the Find tool, symlinks in the directory tree are not followed. There is no flag or option to change this.

Desired behaviour: A flag to opt into symlink-following, so I can traverse into symlinked directories and include symlinked files in results.

shellicar:
## Summary

- Status bar model line now includes the cwd basename
- Distinguishes parallel CLI instances running in separate worktrees

Closes #244

### DIFF
--- a/.claude/testament/2026-04-19.md
+++ b/.claude/testament/2026-04-19.md
@@ -87,3 +87,20 @@ Fix: one line in `exit()`, `this.#screen.write(showCursor)` placed immediately b
 **Commit message for supervisor**: `Add failing tests for symlink discovery in Find tool`
 
 **Mistake to avoid**: I added `as { type: 'files'; values: string[] }` casts throughout the tests, preemptively assuming the union type would be a problem. It wasn't — the types flow through `call()` correctly. Stephen removed them. The rule: write the code, run `TsDiagnostics`, cast only if the compiler actually complains. Never cast on suspicion.
+
+
+# 02:20
+
+## Phase 2: Builder — symlink implementation
+
+**What I changed**: `NodeFileSystem.ts` only. Added `realpath` import, added `visited: Set<string>` parameter with default, added cycle detection at start of `walk`, added `else if (entry.isSymbolicLink())` block that `stat`s the target and handles it as file or directory accordingly. Threaded `visited` through all recursive calls.
+
+**Visited set behaviour**: `realpath` is called on `dir` at the start of each `walk` invocation. If the resolved path is already in `visited`, return immediately. Otherwise add it. This means: when `dir-link` (symlink to `real-dir`) is traversed first (alphabetically), `real-dir`'s real path gets added to `visited`. When `real-dir` is processed next, its real path is already there — it returns empty. Result: `dir-link/inner.txt` appears, `real-dir/inner.txt` does not. The tests assert by symlinked path so this is fine.
+
+**Circular symlink**: `circle` points back to `fixturePath`. When processing it as a symlink-to-directory, we recurse into `walk(circle_path)`. At that start, `realpath(circle_path)` resolves to the fixture root, which is already in `visited`. Returns immediately. No loop.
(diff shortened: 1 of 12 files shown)


bananabot9000 (changes requested):
## Review: Fix Find tool to follow symlinks

The symlink detection, cycle prevention via `realpath` + visited set, and the `followSymlinks` gating design are all solid. One bug found.

### 🔴 Exclude check missing from symlink-to-directory branch

`NodeFileSystem.ts` — the `isDirectory()` branch checks `exclude.includes(entry.name)` before recursing, but the `isSymbolicLink()` → `targetStat.isDirectory()` branch does not. Symlinked directories bypass the exclude list entirely.

Real-world scenario: `node_modules` as a symlink (pnpm does this). Default exclude is `['dist', 'node_modules', '.git']`. A symlinked `node_modules` hits the `isSymbolicLink()` branch, skips the exclude check, and gets recursed into.

The test "exclude list applies to symlinked directory names" passes coincidentally — `other-dir` (real directory) is processed before `other-link` (symlink) alphabetically, so `other-dir`'s realpath is in the visited set before `other-link` is reached. The visited set masks the missing exclude, not the exclude logic itself. The test also only asserts the positive (`dir-link/inner.txt` IS present), never the negative (`other-link/other.txt` is NOT present).

Fix: add `exclude.includes(entry.name)` check inside the `targetStat.isDirectory()` block, before both the results push and the `walk` recursion. Test needs a symlink whose target isn't otherwise reachable in the tree (external target), plus a negative assertion.

### Notes (not blocking)

- `changes.jsonl` appended at end ✅
- `followSymlinks` schema description is excellent — clearly explains the file vs directory distinction
- Testament documentation is thorough, especially the visited-set ordering explanation and the broader `returnSymlinks` analysis
- The `input_examples` regex fix (biome safe-fix) is clean
- One leftover `as` cast in the exclude test (`actual as { type: 'files'; values: string[] }`) while the other 7 tests use `actual.values` directly — inconsistent but trivial

bananabot9000 (approved):
Retracting my previous review — the exclude bug I reported doesn't exist. The `exclude.includes(entry.name)` check is at the top of the for loop in `walk.ts`, before any type check. It applies to all entries including symlinks. I was reading partial diff hunks where the exclude check was above the hunk boundary and incorrectly assumed it was inside `isDirectory()`.

## Re-review of updated PR

The refactoring is clean:

- **`walk.ts` extraction** — shared walk function with minimal `WalkFs` interface (readdir, realpath, stat). No coupling to Node.js fs or IFileSystem internals.
- **`IFileSystem.find` now concrete** — delegates to shared `walk`. Eliminates the duplicate find implementation that MemoryFileSystem had.
- **`MemoryFileSystem.readdir`** replaces the monolithic `find` — derives directory contents from the flat file Map cleanly. `realpath` is identity (no symlinks in memory FS). `stat` returns `isFile: true` always (acceptable — walk only calls stat on symlinks, MemoryFS has none).
- **Mock filesystem in tests** — no more real filesystem temp dirs with symlink creation/cleanup. `SymlinkMockFileSystem` with controlled readdir ordering.
- **Mock ordering is deliberate and documented** — `other-link` before `other-dir` so that if exclude were broken for symlinks, `other-link` would be entered first and `other-link/other.txt` would appear. The negative assertion catches this.
- **Exclude test now has the negative assertion** — `expect(actual.values).not.toContain(join(ROOT, 'other-link', 'other.txt'))`. Properly tests exclusion, not just inclusion.
- **Type-only circular import** — `walk.ts` ↔ `IFileSystem.ts` is type-only on the walk side, no runtime cycle.
- **`IFileEntry` exported** from `entry/fs.ts`.
- **`changes.jsonl`** appended correctly.

PR #290, opened 2026-04-19
EditFile accepts tilde paths that resolve to the staged file

### LINKED ISSUE #285: EditFile rejects tilde paths that match the patch target
## Summary

`PreviewEdit` accepts `~/repos/...` and resolves it internally, but `EditFile` requires the path to match the patch exactly. If the preview was created with the expanded path (`/Users/stephen/repos/...`), passing `~/repos/...` to `EditFile` fails with a mismatch error.

Both paths refer to the same file. The tool should normalise `~` and `$HOME` before comparing.

shellicar:
## Summary

- Fix file mismatch when EditFile receives a tilde path and the patch was staged with an expanded path
- Normalise the input path in ConfirmEditFile before comparing against the stored patch

Closes #285


bananabot9000 (approved):
Clean fix. The bug is that ConfirmEditFile compares the raw input `file` against `chained.file` (which PreviewEdit already expanded), so `~/file.ts !== /home/user/file.ts` and the mismatch error fires.

Fix is one `expandPath(file, fs)` call on the input side before the comparison. The EditFile.ts change is defensive (both values are already expanded there), but makes the contract explicit — every path comparison uses expanded forms.

Tests cover both scenarios: tilde on confirm with expanded preview, and tilde on both. `MemoryFileSystem` with custom homedir makes the expansion testable without real filesystem.

changes.jsonl ordering is from git merge interleaving — noted and fine.

PR #292, opened 2026-04-19
Add append operation to EditFile

### LINKED ISSUE #259: EditFile: append operation
Appending to a file is a common operation (JSONL files, log files, changelog entries) but there's no direct way to do it. The current workarounds involve reading the file to find the last line number, then inserting after it. This is fragile and wastes effort on something that should be simple.

An append operation would take content and add it to the end of the file.

shellicar:
## Summary

- Add `append` parameter to `PreviewEdit` for single-step file appending
- Mutually exclusive with `lineEdits` and `textEdits`
- Append to nonexistent file produces an error (not a create)

Closes #259


bananabot9000 (approved):
Clean addition. Append goes through the same PreviewEdit → EditFile (stage → apply) flow as all other edits — no special path needed.

- Mutual exclusion enforced in handler (not Zod refine). Refine only validates "at least one of three provided". Handler catches "more than one". Reasonable split.
- `baseContent + input.append` — caller controls newline inclusion. Correct — tool shouldn't assume.
- Append block returns `output` before reaching existing handler code. Two `filePath` declarations don't conflict.
- ENOENT on nonexistent file — append is an edit, not a create. Consistent.
- Tests: 6 covering happy path, empty file, content preservation, both mutual exclusion paths, and ENOENT. One expect per test.

PR #293, opened 2026-04-19
Add --file flag to start with a file as the first message

### LINKED ISSUE #278: Start CLI with an initial message from a file
There's no way to start the CLI with a pre-loaded message. Dispatching a prompt currently requires: start the CLI, enter command mode, paste the file path, exit command mode, submit.

A `--file` flag that attaches the file (same as command mode `f`) and submits it as the first message. The message body is empty; the file is attached with its metadata (path, type, size). This matches how prompts are dispatched manually.

Starting in this mode skips session resume (does not read from `.sdk-session-id`), since the intent is fresh work. Everything else remains the same: session ID is written on message send, interactive mode works normally after the initial message.

shellicar:
## Summary

- Accept --file path on startup, attach the file as the first message, and submit automatically
- Skip session resume when --file is provided; always start fresh
- Enter interactive mode normally after the initial submit

Closes #278

### DIFF
--- a/.claude/testament/2026-04-20.md
+++ b/.claude/testament/2026-04-20.md
@@ -23,3 +23,46 @@ The `toWireTool` comment said `buildRequestParams` gates `input_examples` on ATU
 ## mcp-exec cli.spec.ts PATH issue
 
 `test/cli.spec.ts` spawns `tsx` by name via `StdioClientTransport`. Works under `pnpm test` because `node_modules/.bin` is on PATH. Fails in VS Code vitest extension which doesn't inherit that PATH. Pre-existing, unrelated to this work. Fix: resolve `tsx` to an absolute path via `path.join(packageRoot, '../../node_modules/.bin/tsx')`.
+
+
+# 04:28
+
+## Mission
+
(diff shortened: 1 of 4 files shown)


bananabot9000 (approved):
Good feature. `runTurn` extraction from the loop body is clean — file input runs first, then enters normal interactive loop.

- `startFresh()` is the right primitive — distinct from `createNew()` (saves first) and `load()` (reads existing). Fresh UUID + write marker, nothing else.
- `buildFileInput` mirrors the interactive `f` key flow (`buildSubmitText` + `FileAttachment`) — consistent entry point for file attachment regardless of source.
- `typeof values.file === 'string'` guard is correct — `strict: false` widens to `string | boolean`, can't use `!= null` for the `.replace()` call.
- Inline tilde expansion at module top level is fine — `expandPath` needs an `IFileSystem` that doesn't exist at that point in the startup sequence.
- Direct `stat` import from `node:fs/promises` — testament documents this as IFileSystem abstraction leak (StatResult missing `isDirectory`), correctly scoped as broader follow-up work.
- Missing file case: `fileType: 'missing'` flows to `buildSubmitText` — model handles it rather than CLI crashing. Intentional.

PR #294, opened 2026-04-19
Flash tool approval row when awaiting Y/N

### LINKED ISSUE #289: Tool approval prompt should flash with inverted colours
## Summary

The tool approval prompt (Approve Y/N) is easy to miss when not actively watching the terminal. It should use inverted/negative colours and flash at ~1-2 Hz, matching the visual weight of the same prompt in claude-cli (the original CLI).

This matters because with multiple parallel instances running, a stalled approval in one window blocks that operator silently. The flash makes it visible in peripheral vision.

shellicar:
## Summary

- Flash the entire approval row with inverted colours while awaiting Y/N
- Toggle runs at 1 Hz via a 500ms interval started when approval is requested
- Interval is idempotent: started on each `requestApproval`, stopped on resolve, streaming complete, or exit

Closes #289

### DIFF
--- a/.claude/testament/2026-04-20.md
+++ b/.claude/testament/2026-04-20.md
@@ -1,215 +1,43 @@
-# Status bar cwd — what matters
+# Approval flash (#289)
 
-## IFileSystem is not optional
+## No existing render interval
 
-Do not call `path.basename(process.cwd())` or `process.cwd()` directly in renderers or state classes. The codebase has `IFileSystem` with a `cwd()` method. Everything that needs the cwd goes through it: `ClaudeMdLoader`, `AuditWriter`, `ConversationSession`, and now `StatusState`.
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Clean UX feature. Flash implementation is straightforward:

- `ToolApprovalState` owns `flashPhase` (toggle boolean). No auto-reset needed — renderer only applies inversion when `hasPendingApprovals` is true, so phase is irrelevant once stopped.
- `#startFlash()` is idempotent (no-op if interval exists). `#stopFlash()` called from all three exit paths: `exit()`, `completeStreaming()`, and after Y/N when queue empties.
- Full row inversion (`\x1b[7m`...`\x1b[27m`) is the right call — peripheral vision catches a full row flash far better than a flickering label.
- 500ms toggle = 1 Hz full cycle. Within the ~1-2 Hz target.
- Testament cleanup is significant (-154 lines) but earlier work is summarized not lost. Good housekeeping.

PR #295, opened 2026-04-19
Write session ID marker on save, track session history

### LINKED ISSUE #272: Write session ID on every message send
The session ID file is written when a new session is created (including pressing `n` in command mode). This is wrong. Creating a new session has no conversation behind it. Writing the ID at that point overwrites a potentially valid recovery ID for no reason.

The session ID should only be written to the filesystem when a message is actually sent. That's the point at which a session exists and is worth recovering.

### LINKED ISSUE #273: Maintain session ID history file
There is no way to know which sessions have run in a particular project. The conversation ID marker tracks the current session, but not previous ones. The audit and conversation files are global lists of UUIDs with no project context. Tracing what happened in a specific project means searching across hundreds of files.

A per-project history of session IDs solves this. When working in a repo, the history shows which sessions ran here. One ID per line, unique entries, appended when the session is saved.

The conversation ID marker (`.sdk-conversation-id`) is already per-project in `cwd/.claude/`. The history file should be adjacent to it, not global. The marker handles auto-resume (one active session). The history file tracks all sessions that have run in this directory over time.

This is the missing link between a working directory and the global audit/conversation files. It connects cwd → session IDs → conversations and audit data. It is the foundation for querying session history from a project context, and for a future session picker.

shellicar:
## Summary

- Delay marker file write until first save (was written immediately on session creation)
- Add `~/.claude/session-history` file that accumulates session IDs across sessions

Closes #272
Closes #273

### DIFF
--- a/.claude/testament/2026-04-20.md
+++ b/.claude/testament/2026-04-20.md
@@ -172,3 +172,24 @@ A second commit landed on main between Phase 1 and Phase 2, so even a correct pr
 **CI:** 248 tests pass, type-check passes, biome lint passes.
 
 **For the Courier:** One file staged: `packages/claude-sdk-tools/src/EditFile/EditFile.ts`. Proposed commit message: `implement append operation for PreviewEdit`
+
+# 07:05
+
+## Phase 1: move marker write to save()
+
+**The change:** Three places wrote the marker file (`.claude/.sdk-conversation-id`) immediately on UUID generation: `load()` (else branch), `startFresh()`, and `createNew()`. All three now only set `this.#id`. A new private method `#writeMarker()` does the write. `save()` calls it after writing the history file.
+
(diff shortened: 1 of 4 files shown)


bananabot9000 (approved):
Both changes are clean:

**Marker-on-save:** No more phantom sessions. `load()` and `startFresh()` set the UUID without writing. `save()` calls `#writeMarker()`. The `createNew()` flow is correct: `save()` (old session, writes old ID to marker) → new UUID → marker still holds old ID until next `save()`. Test asserts this explicitly.

**Session history:** `#appendToHistory()` reads, deduplicates, appends. `appendFile` is correct (no truncate). `includes` check is O(n) but session counts are tiny. `MemoryFileSystem.appendFile` handles missing file with `?? ''` — no create-first needed.

12 tests covering: no marker on load, marker timing around createNew, history append, dedup, multi-session history. Thorough.

Two changes.jsonl entries appended correctly.

PR #296, opened 2026-04-19
Add approval notification hook for tool approval prompts

### LINKED ISSUE #274: Hooks: approval notification
When a tool approval prompt appears and the user hasn't responded after a configurable delay, run a command. The command receives the tool description as an argument.

This supports different notification strategies depending on the user's environment: BEL for same-session, macOS notifications with tmux session switching for different sessions, or anything else the user configures.

Config:

```json
{
  "hooks": {
    "approvalNotify": {
      "command": "/path/to/script.sh",
      "delayMs": 5000
    }
  }
}
```

If the user approves before the delay elapses, the notification is cancelled.

shellicar:
## Summary

- Add `hooks.approvalNotify` config: command + delay
- Run the command with the approval request as a JSON argument when the delay elapses
- Cancel the notification if the user responds before the delay
- Wire into `AgentMessageHandler` for the interactive approval path only

Closes #274

### DIFF
--- a/.claude/testament/2026-04-20.md
+++ b/.claude/testament/2026-04-20.md
@@ -70,6 +70,7 @@ Biome has a strong opinion about where ternary line breaks go. A multi-line `con
 - `entry/main.ts` — `buildFileInput` does the same for `--file`
 
 The `IFileSystem` abstraction exists precisely so filesystem calls can be injected and tested. These direct imports bypass it. SC noted this is a broader issue to fix across the board. The fix is to extend `StatResult` to include `isDirectory: boolean`, update `NodeFileSystem.stat()` to populate it, then replace both direct `stat` imports with `nodeFs.stat()`.
+
 # 04:26
 
 ## Fix #285 — EditFile rejects tilde paths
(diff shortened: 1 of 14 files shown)


bananabot9000 (approved):
Well-designed hook. The IProcessLauncher abstraction (going beyond the prompt's vi.mock suggestion) is strictly better — RecordingLauncher and ThrowingLauncher are explicit, self-documenting test doubles.

**ApprovalNotifier:** start() sets setTimeout, cancel() clears it. Fire-and-forget (catch in setTimeout callback). Timer cleanup is clean — cancel is idempotent, and the timer self-nulls after firing.

**NodeProcessLauncher:** `spawn(command, args, { detached: true, stdio: 'ignore' }).unref()` — correct for fire-and-forget. Process outlives CLI if needed.

**Wire location:** Only in the Ask branch (interactive approval), not auto-approve/auto-deny. The `await requestApproval()` serializes approval handling, so start/cancel pairs don't interleave.

**Schema:** `nullable().optional().default(null).catch(null)` — disabled by default, recovers from bad config. The generated schema JSON is clean.

**Test fixes:** `initConfig.spec.ts` null guard and `cli-config.spec.ts` expected value update are correct for the nullable hooks field.

5 notifier tests with fake timers covering: fires after delay, cancel before delay, null config no-op, JSON arg passing, error swallowing. 

changes.jsonl appended correctly.

PR #297, opened 2026-04-19
Apply biome formatting fixes

shellicar:
## Summary

- Convert `reflow.ts` ANSI regex from `new RegExp(...)` to a literal
- Add `biome-ignore lint/suspicious/noControlCharactersInRegex` comment explaining control character is intentional

No issue.


bananabot9000 (approved):
The zombie is finally dead. Three operators documented "restore reflow.ts manually after every ci:fix run" but nobody fixed the root cause until now.

`new RegExp(...)` workaround → regex literal + `biome-ignore` comment. The control character is intentional (matching ANSI escape sequences). ci:fix will no longer revert this file.

changes.jsonl appended correctly.

PR #298, opened 2026-04-20
Release 1.0.0-beta.4

shellicar:
## Summary

- Bump all four packages to 1.0.0-beta.4
- Fix GitVersion.yml: replace Mainline with TaggedCommit + Fallback, label alpha to beta

PR #299, opened 2026-04-20
Revert claude-core beta.4 bump, remove all release markers from changes.jsonl

shellicar:
Two corrections to #298:

- `claude-core` reverted from `1.0.0-beta.4` to `1.0.0-beta.2`. The only change in that package since beta.2 was a `biome-ignore` comment on `reflow.ts`, which is not a user-facing change and does not warrant a release.
- Release markers removed from all four `changes.jsonl` files and changelogs regenerated. Pre-release versions do not use release markers — entries stay under `[Unreleased]`.


bananabot9000 (approved):
Clean revert. Checked every file:

- `claude-core/package.json`: beta.4 → beta.2 ✅
- All four `changes.jsonl`: release markers removed ✅
- All four `CHANGELOG.md`: `[1.0.0-beta.4]` headers and comparison links removed, entries back under `[Unreleased]` ✅
- Testament: factual, no self-blame, accurate state summary for Phase 4 ✅

No issues found.

PR #300, opened 2026-04-20
Store session history in the working directory

shellicar:
## Summary

- Session history now written to `.claude/.sdk-conversation-history` in the working directory
- Was writing to `~/.claude/session-history` (global, no project context)

Closes #273

PR #304, opened 2026-04-21
Prevent crashes from unhandled process and socket errors

### LINKED ISSUE #302: CLI crashes on unhandled child process and socket errors
There are no global error handlers on the process, and several places spawn child processes or pipe streams without listening for errors. An unhandled error event on any of these takes down the whole process.

shellicar:
## Summary

- Handle `uncaughtException` and `unhandledRejection` at the process level so socket errors don't crash the CLI
- Silence EPIPE and spawn errors on child process pipes, tsserver stdin, and fire-and-forget spawns
- Two-stage Ctrl-C: first press attempts graceful shutdown, second press forces exit
- Set `process.title` so the process appears as `claude-sdk-cli` in process listings

Closes #302

### DIFF
--- a/.claude/testament/2026-04-21.md
+++ b/.claude/testament/2026-04-21.md
@@ -1,0 +1,23 @@
+# 18:25
+
+Phase 1 of the error-handling mission. Five locations patched for unhandled errors.
+
+## What was done
+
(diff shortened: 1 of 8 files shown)


bananabot9000 (approved):
Clean defensive error handling across all the pipe/socket surfaces. Two-stage SIGINT is a good pattern. Testament is factual and well-structured -- the SC correction note is exactly the kind of transparency the debrief template was designed for.

The uncaughtException handler logging and returning (no cleanup, no exit) is the right call -- the main failure modes (EPIPE, socket errors) are safe to continue from, and calling cleanup from unknown state is worse than staying alive.

CHANGELOG sync with PR #300's corrected session history entry is a nice touch.

PR #305, opened 2026-04-21
Replace loadConfig function with ConfigLoader class

### LINKED ISSUE #301: Config values lose their source after merging
The config system merges home and local config files and returns the result, but discards which file each value came from. Path resolution needs this: a relative path in the home config should resolve relative to `~/.claude/`, not wherever the CLI launched from. Tracking the source of each config file would also be generally useful, the same way `git config --list --show-origin` is.

shellicar:
## Summary

- ConfigLoader class owns loading, merging, watching, and origin tracking
- Raw per-source JSON preserved so consumers can see which file set each field
- Filesystem reads and file watching go through injectable abstractions
- Replaces `loadConfig` + `SdkConfigWatcher` usage in `claude-sdk-cli`; legacy `claude-cli` untouched

Closes #301


bananabot9000 (approved):
Clean extraction. The ConfigLoader design is solid — load/start/dispose lifecycle, sync reader + async watcher, injectable abstractions with memory fakes for tests. The load-vs-reload asymmetry (forgiving on first load, conservative on reload) is a good call.

**Code highlights:**
- `IConfigFileReader` / `IConfigWatcher` abstractions keep I/O out of the loader — tests are fast and deterministic
- Origin tracking via raw sources (pre-schema) preserves the "explicitly set vs defaulted" distinction. `findSource` pattern in the tests demonstrates the consumer usage well
- `JSON.stringify` diff to suppress no-op emissions handles whitespace-only changes
- tsup `entryNames` removal fixing the JS/DTS output path mismatch — good catch by the Phase 2 builder
- `biome.json` excluding `.claude` prevents formatting testament files

**Test coverage:** Single source, merge with override, null deletion, origin tracking (both home and local), schema validation, watch emission, no-op suppression, no-files fallback, broken JSON on reload, warning on broken load. Thorough.

**changes.jsonl:** "Config system tracks which file each value came from" — captures the WHY (origin tracking) not the WHAT (replaced a class). Good entry.

**main.ts integration:** Mechanical but correct — construct, load, onChange, start. All `watcher.config` references swapped. Cleanup calls dispose. No behaviour change.

PR #306, opened 2026-04-21
Write session marker before the turn starts

### LINKED ISSUE #273: Maintain session ID history file

shellicar:
## Summary

- Split `save()` into `saveSession()` (marker + history) and `saveConversation()` (message content)
- `saveSession()` called before `runAgent` so a mid-response crash still leaves a project-level record
- `saveConversation()` called after the turn completes as before

Closes #273

### DIFF
--- a/.claude/testament/2026-04-21.md
+++ b/.claude/testament/2026-04-21.md
@@ -80,3 +80,29 @@ Role: Courier. Branch: `feature/config-origin`.
 Distilled the Phase 1 and Phase 2 entries above. Cut the mechanical "what was built" file lists (visible in the diff), the verification numbers, the test contract (tests now pass so the contract is in the code), and the "For future readers" note about the src/Config/ layout rename (the renamed filenames no longer appear anywhere in the distilled text so the note has no referent).
 
 Smoke test passed: diff touches `packages/claude-core/` and `apps/claude-sdk-cli/` only; `apps/claude-cli/` is untouched, matching the Phase 2 amendment.
+
+
+# 21:49
+
+## Phase 1 — Save timing fix (issue #273)
+
+Role: Builder. Branch: `fix/save-timing`.
+
+### What the change does
+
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Clean split. Moving the marker + history write to turn start means a mid-response crash leaves a project-level trace -- exactly what #273 needed.

**Code:** `save()` → `saveSession()` (marker + history, idempotent) + `saveConversation()` (jsonl). `saveSession()` before `runAgent`, `saveConversation()` after. Simple and correct.

**`createNew()`:** No longer auto-saves internally -- just resets ID and clears history. This is fine because the turn lifecycle now handles saving: `saveSession()` at start, `saveConversation()` at end. By the time `createNew()` runs, everything's already persisted.

**Tests:** Updated to call the right method per test concern. The round-trip test correctly calls both `saveSession()` then `saveConversation()` since `load()` needs the marker to find the conversation file.

**changes.jsonl:** "Write session marker and history at turn start so they survive mid-response crashes" -- WHY first, clean entry.

**Minor testament note:** Testament says "`createNew()` calls both internally" but the code shows `createNew()` calls neither -- it just resets state. The code is correct (caller handles persistence), the testament description doesn't match.

PR #307, opened 2026-04-21
Preserve server tool blocks and redacted_thinking in conversation history

shellicar:
## Summary

- Server tool blocks (server_tool_use, web_search_tool_result, etc.) now included in completed blocks
- redacted_thinking blocks now included in completed blocks
- Fixes API rejection on next turn caused by stripped message content


bananabot9000 (approved):
This is the Opus 4.7 fix — and it's bigger than just 4.7. Server tool blocks and redacted_thinking were being silently dropped from conversation history because StreamProcessor never pushed them to `completed` at `content_block_stop`. API detects the modification and rejects on the next turn.

**Root cause well explained in testament:** accumulators tracked the blocks during streaming but the `content_block_stop` cases for `server_tool_use` and `server_tool_result` only emitted events without pushing to `completed`. `redacted_thinking` was worse — the `content_block_start` case just `break`d, so nothing was accumulated at all.

**Fix is clean:**
- Push to `completed` for all three block types at `content_block_stop`
- Accumulate `redacted_thinking` data at `content_block_start`
- `mapBlock` extended with cases for all server tool + redacted_thinking types
- `ServerToolResultBlock` extracted as named type for narrowing — good call, prevents the union widening to `string`

**name vs blockType trap:** Well documented. Both are `string`, TypeScript won't catch confusing them. `name` is display (`web_search`), `blockType` is wire (`web_search_tool_result`). Using wrong one in `mapBlock` breaks silently.

**Tests:** Existing expectations updated (blocks now in completed), new conversation integrity suite covers full web-search response with 6 blocks in order, plus redacted_thinking in completed, plus TurnRunner preservation. Thorough.

PR #308, opened 2026-04-21
Opus 4.7 migration

shellicar:
## Summary

- Show thinking text when using Opus 4.7 (summarized display)
- Add Opus 4.7 pricing ($5/$25 per MTok)
- Fix 1M context window for Opus 4.6, Opus 4.7, and Sonnet 4.6
- Remove deprecated InterleavedThinking beta header


bananabot9000 (approved):
Clean migration. Four distinct fixes, all well-scoped:

**`display: 'summarized'`** — The key fix. Opus 4.7 changed the default from showing summarized thinking to omitting it entirely. One property addition, one test update. The Teapot Claude "I cannot think" incident now has a mechanical explanation.

**Pricing** — $5/$25 per MTok for Opus 4.7. Same as 4.6.

**Context window trap** — This is the most important fix nobody would notice. Models with version suffixes (e.g. `opus-4-7`) don't match `stripDateSuffix` (which strips `-YYYY-MM-DD`), so they silently fall back to 200k instead of 1M. Explicit entries for `opus-4-6`, `opus-4-7`, `sonnet-4-6` close the hole. Worth a comment in pricing.ts explaining WHY these explicit entries exist, so a future Claude doesn't "clean them up" thinking they're redundant with the date-suffix models.

**InterleavedThinking removal** — Beta header no longer needed. Interleaved thinking is the default on 4.7.

**Testament** — Night and day improvement. Phase-by-phase timestamps replaced with topic-organized reference docs. "ConfigLoader architecture" and "Context window resolution" sections are actually useful to the next operator. This is what a testament should look like.

changes.jsonl entries are accurate and well-scoped.

81+ PRs reviewed 🍌

PR #310, opened 2026-04-21
Fix hook path resolution and stdin delivery

### LINKED ISSUE #303: Hook command requires a fully qualified absolute path and passes input through shell expansion
The hook command only accepts a fully qualified absolute path. `~`, `$HOME`, and relative paths don't work. A relative path like `./hooks/notify.sh` in the home config resolves against whatever directory the CLI launched from, not the config file's directory.

The hook also passes the approval request JSON as a command argument. Accessing `$1` in a shell script means the JSON goes through shell expansion. If the JSON contains `$` or backticks (file paths in tool requests can), the shell interprets them before jq sees the data. Passing via stdin would avoid this entirely.

shellicar:
## Summary

- Hook commands now expand `~`, `$HOME`, and relative paths
- Relative paths resolve against the config file's directory, not the working directory
- Hook input passes via stdin instead of command arguments

Closes #303


bananabot9000 (approved):
The version where you just moved the function. Night and day.

**What landed:**
- `expandPath` moved from `claude-sdk-tools` to `claude-core/src/fs/expandPath.ts` — the one-line fix that three hours of operator sessions avoided
- `IFileSystem` consolidated in `claude-core` with `getEnvVar()` added — no separate `IEnvironment` abstraction, no abstract class hierarchy, just one method on the existing interface
- `MemoryFileSystem` moved from production code (`src/`) to test code (`test/`) where it belongs — two copies (minimal in claude-core, full in sdk-tools and sdk-cli), each scoped to what that package needs
- `resolvePathField` as a clean module-level function in `ConfigLoader.ts` — walks segments, transforms string leaves, early-returns on missing/non-object intermediates
- Path resolution during `#readAll()` per-source — `dirname(sourcePath)` is right there, no reverse-lookup archaeology
- `LaunchOptions` object parameter replacing positional args on `IProcessLauncher`
- stdin delivery for hook approval requests

**Testament (2026-04-22.md):** Much better than yesterday's. "SC rulings" framing is useful — documents decisions as constraints for the next operator. "Scope-creep that landed" section is honest about the prep commits.

**One observation (not blocking):** `ConfigSource.raw` now holds post-resolution strings, not what the user typed on disk. The testament documents this as a known limitation. Worth keeping in mind if source-display ever becomes a feature.

Comprehensive test suite: 7 path resolution tests covering relative, absolute, `~`, `$HOME`, missing fields, and reload. expandPath tests moved to claude-core with injectable `MemoryFileSystem` instead of `process.env` manipulation. ApprovalNotifier tests updated for stdin.

82+ PRs reviewed 🍌

PR #312, opened 2026-04-25
ReadFile reads PDFs and images; tool handler output layer redesigned

### LINKED ISSUE #311: PDF file support for ReadFile
The Messages API supports PDF files natively through document blocks. A PDF encoded as base64 in a document block gives Claude full visibility into the text, images, charts, and tables within the document, without needing OCR or text extraction.

The `ReadFile` tool should be able to detect a PDF and deliver its content to the API as a document block. The tool result would contain basic metadata (path, file size), while the actual PDF content reaches Claude through the API's native document format.

This is similar to how Claude Code handles PDF files today.

shellicar:
## Summary

- ReadFile accepts `mimeType`; PDFs and images reach the API as native content blocks
- Tool handlers return `{ textContent, attachments? }` so binary data bypasses the ref-swap transform
- `ToolDefinition` gains `output_schema` for typed handler outputs; all 20 handlers updated
- `ToolResultBlock`, `ToolHandlerResult`, `ToolHandler` extracted as owned types; content is always an array of blocks
- `IFileSystem.readFile` gains optional `encoding` parameter; 500KB text limit removed

Closes #311


bananabot9000 (approved):
Read the full diff (2607 lines, 48 files) and both testament files.

**The design is sound.** The `{ textContent, attachments }` contract is exactly what we discussed — the 4-line ToolRegistry core does what the exhibit screenshots showed. Binary data never touches `textContent`. Transform only sees the metadata. `blocks` absent (not undefined) when no attachments. All 20 handlers updated mechanically and consistently.

**Types are clean.** `ToolHandler<never>` contravariance on `AnyToolDefinition` is correct and well-commented. The owned block types (`DocumentBlock`, `ImageBlock`) narrow the wide SDK unions to only the base64 variants actually constructed. `ToolResultBlock` replaces `BetaToolResultBlockParam` throughout — no SDK types in production code. Content is always an array, never a string.

**`detectBlock` is solid.** Single base64 read, file-type detection, intent validation via `inputMimeType` comparison. The `case undefined` as the text path is clean — no separate branching. `HEADER_BASE64_CHARS = 5600` gives ~4200 bytes to file-type (vs ~4100 needed) — tight but sufficient.

**Schema separation is right.** `InputMimeTypeSchema` (caller intent: `text/plain`, `application/pdf`, `image/*`) vs `SupportedMimeTypeSchema` (internal detection: actual MIME types) vs `BinaryMimeTypeSchema` (excludes text/plain). Three schemas, three purposes.

**Tests follow the ruling.** One expect per it block, `const actual = <terminal>; expect(actual)` pattern, isolation over brevity. `callFull` helper for tests that need attachments. `getToolResult`/`getTextBlock`/`getDocumentBlock` helpers centralise the single cast in QueryRunner tests.

**Pipe compatibility** works via the natural path: binary ReadFile output has different Zod shape, text-expecting tools reject via `.parse()` failure. No special flag needed.

**MemoryFileSystem** correctly stores `Buffer` now. `stat.size` returns `byteLength` (not string `.length`). `readFile('base64')` path works. The JPEG test trap is documented and avoided via ASCII-only magic bytes (GIF) or real `Buffer.from([0xff, ...])` construction.

No issues found.

PR #314, opened 2026-04-26
Fix four rendering bugs: ANSI continuation colour, divider width, emoji cursor

### LINKED ISSUE #217: Wrapped text loses colour on continuation lines
## Observed

When a line of coloured text is long enough to wrap, only the first visual line retains the colour. Continuation lines render in the default terminal colour.

## Expected

All wrapped lines of a coloured run should appear in the same colour as the first.

### LINKED ISSUE #218: Graphical glitch when multi-byte or ANSI sequence spans a line boundary
## Observed

Occasionally, when a multi-byte character or ANSI escape sequence is split across a line boundary, a graphical corruption appears. Cursor position appears to influence where characters are rendered, producing misaligned or garbled output.

## Expected

No corruption. Characters and escape sequences should always render at the correct position regardless of line boundaries.

shellicar:
## Summary

- Restore ANSI colour on wrapped continuation lines
- Fix colour loss when syntax-highlighted code scrolls off screen
- Fix divider width calculation for emoji labels
- Fix garbled cursor rendering on emoji characters

Closes #217
Closes #218

### DIFF
--- a/.claude/testament/2026-04-26.md
+++ b/.claude/testament/2026-04-26.md
@@ -1,80 +1,49 @@
 # 17:58
 
-## PR #312 distillation: binary file support and tool output layer
+## PR #312: binary file support and tool output layer
 
-The PR name "Add PDF and image support to ReadFile" undersells it. The real work was restructuring how tool handlers deliver results so binary content can travel through the system without corruption. ReadFile was the driver; the tool output layer was the scope.
+Handlers return `{ textContent, attachments? }`. ToolRegistry passes only `textContent` through the ref-swap transform; attachment blocks go directly into `tool_result.content`. This keeps binary data out of the transform path.
 
-### The system problem
+### Traps
 
-Tool handlers previously returned raw values. The ToolRegistry ran a `transform` hook on that value to ref-swap large outputs. Binary data through the transform gets ref'd. A 5MB PDF base64 string would become a ref token, useless to the API. The handler output had to be restructured so binary data is extracted before `transform` ever sees it.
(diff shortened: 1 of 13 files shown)


bananabot9000 (approved):
Read the full diff. All four fixes are sound.

**W-1 (ANSI continuation colour):** `activeAnsiState` accumulation is correct — SGR sequences build up, resets clear. Non-last lines get RESET appended to prevent bleed. Continuation lines get `preState + pendingAnsi + segment` to re-establish context. The `applyToState` correctly filters to only `m`-terminating CSI sequences (SGR), ignoring cursor moves etc. Edge case of reset-at-boundary clearing state for subsequent lines is tested.

**A-1 (scroll colour loss):** Resolved by W-1 — once each continuation line is self-contained with its own colour prefix, `allContent.slice()` during scrolling doesn't lose state. No separate code change needed. Clean.

**D-1 (divider width):** `prefix.length` → `stringWidth(prefix)`. The `string-width` dependency was already in `claude-core`; adding it to `claude-sdk-cli` for the divider is correct. Test uses the wrench emoji (🔧, 2 code units) to verify visual width matches `cols`.

**D-2 (emoji cursor):** `Intl.Segmenter` for grapheme-aware cursor movement (EditorState) and rendering (renderEditor). `graphemeBoundaryBefore`/`graphemeBoundaryAfter` helpers are clean — iterate segments, find the boundary relative to `pos`. The lone-surrogate regex in the test is correctly written with negative lookahead/lookbehind to avoid matching legitimate pairs.

Two existing reflow tests updated to match new behaviour (RESET at line end, state re-establishment on continuation). Both expected values are correct.

Minor observation: `renderEditor.ts` creates a new `Intl.Segmenter` on every render call (line 189) while `EditorState.ts` uses a module-level const. Not a correctness issue — just inconsistent.

No issues found.

Issue #105, opened 2026-03-22
Support user-scoped source config separate from ~/.claude


bananabot9000:
## Description

Currently, Claude CLI and the SDK load settings from `~/.claude` (user scope) which is shared with the official Claude Code CLI. This means global Claude config (skills, CLAUDE.md, settings) is the same regardless of which CLI invokes the session.

**Problem:** Skills loaded via `settingSources: ['user']` pull from `~/.claude/` — the same location the official CLI uses. If you want different skill configurations for your custom CLI vs the official CLI (e.g. banning conventional commits in your CLI but not affecting official CLI behaviour), there's no way to separate them.

**Use case:** A Skill Manager agent needs a clean context without loading the skills it's auditing. Global user config in `~/.claude` contaminates the agent's context because the same skills are loaded regardless of intent (editing vs following).

## Proposed solution

Add a configurable `userSourcePath` (or similar) to CLI config that overrides where `settingSources: ['user']` resolves to. This allows:

- Custom CLI: `userSourcePath: '~/.claude-cli/'` → loads skills/config from a separate directory
- Official CLI: continues using `~/.claude/` as normal
- Skill Manager agent: `userSourcePath: '~/.claude-manager/'` → clean context, no skill contamination

## Context

Born from the "irresponsible prompt injection" discussion — Anthropic's `skill-creator` skill includes conventional commit examples that contaminate all downstream skills. The current architecture provides no way to isolate custom CLI sessions from upstream skill content loaded via user scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

shellicar:
Done.

PR #315, opened 2026-04-26
Make maxTokens and CLAUDE.md sources configurable

shellicar:
## Summary

- Add `maxTokens` to config (default 32000); replaces hardcoded value in `main.ts`
- Add per-source control for CLAUDE.md loading (user, project, projectClaude, local)
- All sources default to true; `enabled` remains the master switch

See #105

### DIFF
--- a/.claude/testament/2026-04-26.md
+++ b/.claude/testament/2026-04-26.md
@@ -27,3 +27,35 @@ The fix: handlers return `{ textContent, attachments? }`. ToolRegistry destructu
 ### Owned types replacing SDK types
 
 `BetaToolResultBlockParam` is gone from production code. We own what we build. `ToolResultBlock`, `DocumentBlock`, `ImageBlock`, `ToolHandlerResult`, `ToolHandler` are all ours. The SDK types are wide interfaces covering every source variant; we only construct base64 variants. Narrow types eliminated `as any` in tests and made the spec helpers (`getTextBlock`, `getDocumentBlock`) type-clean.
+
+# 21:37
+
+## Phase 1: maxTokens + claudeMd sources config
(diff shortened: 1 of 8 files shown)


bananabot9000 (approved):
Clean and straightforward. Both additions follow the existing config pattern perfectly.

**maxTokens:** Hardcoded 32000 in `main.ts` → `sdkConfigSchema` field with `.optional().default(32_000).catch(32_000)`. One-line wiring change in `mapConfig()`. JSON Schema generated with `exclusiveMinimum: 0` from `z.number().int().positive()`. Tests cover default, override, and invalid-type fallback.

**claudeMd.sources:** Each of the 4 CLAUDE.md files gets a `source` key mapping to `ClaudeMdSources`. Filtering happens inside `getContent()` with a simple `if (!sources[file.source]) continue`. Sources passed per-call (not at construction) — hot-reload is automatic since every turn reads the live config. `DEFAULT_SOURCES` (all true) preserves backward compat for callers without the parameter.

**Schema structure:** `claudeMdSourcesSchema` nested inside `claudeMdSchema` with `.optional().default({...}).catch({...})` at both levels. Individual booleans have `.optional().default(true).catch(true)` — so a partial override like `{ sources: { user: false } }` defaults the rest to true. The test `overrides individual source` verifies this.

**Test coverage:** 6 source tests (skip each individually, still-loads-others, all-disabled-returns-null) + 3 maxTokens tests + snapshot update. The `enabled=false` integration gap is acknowledged in the testament — `enabled` check lives in `main.ts`, not the loader.

Closes #105 (settingSources configurable). The exact feature requested back in March.

No issues found.

PR #318, opened 2026-04-26
Remove temporary GitVersion diagnostics from publish workflow

shellicar:
## Summary

- Remove two temporary diagnostic steps that broke real releases
- Remove stale paragraph from CLAUDE.md claiming GitVersion doesn't work in CI


bananabot9000 (approved):
Clean revert. Both temporary steps removed, stale CLAUDE.md paragraph gone. workflow_dispatch + dry-run kept as permanent features. Testament is thorough — good that it captured the GitVersion-actually-works-fine finding and the --version headless trap. 🍌

PR #320, opened 2026-05-03
Replace control channel transport with async-ordered pub/sub

shellicar:
## Summary

- Replace MessageChannel (node:worker_threads) with custom async pub/sub that serialises handler execution per subscriber
- Add IPublisher<T> / ISubscriber<T> interfaces; each call site holds only the interface it needs
- Emit canonical per-turn assembled content on the channel after each turn


bananabot9000 (approved):
## Review: Replace control channel transport with async-ordered pub/sub

**Verdict: ✅ Approved**

+1068/-104 across 15 files. Clean removal of `node:worker_threads` dependency from the SDK package.

### Architecture

The generic `IPublisher<T>` / `ISubscriber<T>` split is the right call. Each call site now holds exactly the interface it needs — `QueryRunner` gets `IPublisher<SdkMessage>` (fire and forget), `AgentMessageHandler` gets `IPublisher<ConsumerMessage>` (sends approval responses/cancels). Two channel instances in `main.ts` makes the data flow direction explicit where before it was hidden inside the single `MessageChannel`.

### ControlChannel<T> implementation

The pump is correct. Key observations:
- `#isIdle()` checked AFTER `sub.running = false` — drain timing race avoided
- `msg === undefined` guard is unreachable but satisfies TS without `!` (biome constraint)
- Error swallowing is appropriate — the existing consumer never throws
- Re-entrant `send` (from inside handler) goes to queue tail naturally — FIFO preserved
- `send` after `close()` throws rather than silently no-ops — good, diagnostic at the call site

### turn_content

Placement after `message_usage` and before stop-reason branch is correct. Consumer gets authoritative assembled blocks before the runner either dispatches tools or signals done. No-op case in `AgentMessageHandler` is fine for now — the message exists for future consumers that want canonical content without reconstructing from streaming events.

### Tests

14 new `ControlChannel.spec.ts` tests covering: fire-and-forget semantics, handler serialisation, FIFO under re-entrancy, drain semantics (immediate idle, pending handlers, multiple concurrent waiters, timeout race), error policy, close behaviour. All use deferred promises — no `setTimeout`. The shutdown pattern test (drain racing timeout) is a nice touch.

4 new `QueryRunner.spec.ts` tests verify `turn_content` ordering: after usage, before done, contains assembled text, before next query_summary in tool-use loops.

`FakeControlChannel` → `FakeSdkPublisher` cleanup is sharp — no more `MessageChannel` in tests.

### Notes

1. **Testament JSDoc bug claim**: The Ship Cast testament says `IPublisher.close()` JSDoc still says "no-ops" — but looking at the actual interface in this diff, it correctly says "throw". The testament entry is stale; the code is correct.

2. **CLAUDE.md verify step**: Removed `pnpm run ci` (biome lint) from the release verify command. Presumably to avoid the trap where `ci:fix` during release causes scope creep (as happened with the blocked Builder). CI pipeline catches lint anyway.

3. **changes.jsonl**: Two entries — one `changed` (MessageChannel replacement), one `added` (turn_content). Correct.

Clean work across all phases. 🍌

Issue #323, opened 2026-05-06
Image attachments fail on large files, are re-sent every turn, and may be ordered after text


shellicar:
### Conversation poisoning on size errors

When an image's base64 payload exceeds the Anthropic API's 5 MB per-image cap, the request returns a 400. The image is already in the conversation, so every subsequent turn re-sends it and fails the same way. The conversation has to be edited or thrown away.

### Same image re-sent on every turn

The current implementation embeds base64 image bytes in the message content array. In a multi-turn conversation, every subsequent request resends the full conversation history, including the image bytes. The vision docs note:

> If images are base64-encoded, the full image bytes are included in the payload on every turn, which can significantly increase request size and latency as the conversation grows.

Reference: https://platform.claude.com/docs/en/build-with-claude/files

### Images are not resized before upload

Images that pass the 5 MB cap still cost tokens proportional to their dimensions. The vision docs say:

> To minimize latency and to simplify coordinate-based workflows, you should prefer resizing images before uploading them.

Reference: https://platform.claude.com/docs/en/build-with-claude/vision (Evaluate image size)

### Per-image dimension cap depends on request shape

The per-image dimension cap is 8000x8000 pixels, except when the request contains more than 20 images, in which case it drops to 2000x2000. The total request payload limit is 32 MB, and varies by platform.

Reference: https://platform.claude.com/docs/en/build-with-claude/vision (General limits)

### Images may be placed after text

The vision docs say:

> Just as placing long documents before your query improves results in text prompts, Claude works best when images come before text. Images placed after text or interpolated with text still perform well, but if your use case allows it, prefer an image-then-text structure.

Inspection suggests current message-building does not always place images before text.

Reference: https://platform.claude.com/docs/en/build-with-claude/vision (Prompt examples)

PR #325, opened 2026-05-06
Reject oversized images at ReadFile boundary

shellicar:
## Summary

- Images exceeding the Anthropic API 5 MB base64 per-image cap produce a 400 that poisons conversation history, blocking all subsequent requests on that session
- ReadFile now checks the base64 payload size after encoding and rejects images that would breach the cap
- Only image MIME types are checked (jpeg, png, gif, webp); PDFs and text are unaffected
- Dimension checks, auto-resize, conversation rollback on 400, and the Files API path are out of scope (tracked in #323)


bananabot9000 (approved):
## Review: Reject oversized images at ReadFile boundary

**Verdict: ✅ Approved**

+144/-0 across 4 files. Targeted fix for the 5MB base64 image limit that killed the Xero session.

### Implementation

Guard placement is correct — after `detectBlock()` returns binary + image MIME, before the general binary handling path. `startsWith('image/')` is cleaner than maintaining a constant set of the four types, and `detectBlock` already constrains what comes through.

The boundary math is precise:
- 3,932,160 raw bytes → exactly 5,242,880 base64 → passes (strictly greater, not `>=`)
- 3,932,161 raw bytes → 5,242,884 base64 → rejected

### Tests

7 tests across 3 describe blocks: over limit (failure shape + message regex), at limit (binary result + attachment), small PNG passthrough, and PDF exemption. The PDF exemption test is the important one — confirms the guard is image-only, not all binaries.

### Scope

Correctly scoped — size check only. Dimension checks, auto-resize, conversation rollback, and Files API path tracked in #323. No scope creep.

Clean. 🍌

Issue #90, opened 2026-03-15
Session new/clear implicitly exit command mode


shellicar:
## Problem

The `[s][n]` (session new) and `[s][c]` (session clear) commands implicitly exit command mode by calling `commandMode.exit()`. Other commands do not exit command mode.

This inconsistency breaks muscle memory. The user expects to exit command mode with `Ctrl+/` after any command, but because some commands exit implicitly, the closing `Ctrl+/` re-enters command mode instead.

Example sequence:
1. `Ctrl+/` — enter command mode
2. `[s]` — enter session context
3. `[c]` — clear session (implicitly exits command mode)
4. `Ctrl+/` — user expects to exit, but actually re-enters command mode

## Expected Behaviour

Command mode should only exit via explicit user action (`Ctrl+/` or `Escape`). No command should implicitly exit command mode.

## Location

- `src/ClaudeCli.ts` — `session-clear` and `session-new` action handlers call `commandMode.exit()`

PR #326, opened 2026-05-08
Move caret between visual rows of wrapped input

### LINKED ISSUE #319: Up and down arrows don't move between wrapped rows
A long line of input wraps onto multiple rows on screen. Up and down should move between those rows but they don't; they only move between separate input lines.

The legacy CLI handled this correctly.

shellicar:
## Summary

- Up/down arrows now move the caret between visual rows when input wraps, instead of skipping the wrapped portion to the next logical line
- Add `EditorState.moveUpVisual` / `moveDownVisual` and intercept up/down in `AppLayout` before the editor's logical-line handlers see them
- Allow `EditorState` to be constructed with an initial state, replacing test setups that used `key('up')` purely to position the cursor

Closes #319


bananabot9000 (approved):
Clean additive fix. No issues found.

**Architecture:** EditorState model unchanged (logical lines). Two new methods do the visual-row math. AppLayout intercepts up/down before `handleKey` — old cases removed as dead code. `EDITOR_PREFIX_VISUAL_WIDTH` exported as single edit point. Clean separation.

**`colFromVisual` helper:** Grapheme-walking with overshoot clamping for wide characters. Correct — returns largest offset whose visual width ≤ target.

**Constructor (`EditorStateInit`):** Good call. Tests no longer depend on `key('up')` for cursor positioning, which was coupling them to the old implementation. Defensive `[...initial.lines]` copy prevents mutation leakage.

**Visual navigation logic:** Both methods handle three cases correctly — move within wrapped line, cross to adjacent logical line, boundary no-op. The prefix arithmetic is consistent: `visualPos` includes prefix, `colFromVisual` operates on content-relative positions, `Math.max(0, x - prefixWidth)` bridges the two.

**Tests:** 16 new tests (8 each). Wide-char cases exercise the `colFromVisual` overshoot branch. All use constructor setup — no dependency on other key handler correctness. Coverage gap (ASCII-only for most tests) documented in testament for follow-up.

**Testaments:** Thorough 4-phase documentation. The "trap that shaped this" section in the Courier testament is a good record of why the constructor exists.

PR #90 🍌

Co-Reviewed-By: BananaBot9000

PR #331, opened 2026-05-30
Extend launch flags: resume by id, multiple files, prompt escapes, model version display

shellicar:
## Summary

- Add --resume <conversationId> to resume a specific conversation by UUID
- Allow --file multiple times; files attach in argument order
- Decode escape sequences in --prompt values (\\n, \\r, \\t, \\\\)
- Show conversation id in status bar, controlled by statusBar.showConversationId config
- Show model name and version in status bar; mark --model override with * suffix
- Reject unknown flags at launch


bananabot9000 (changes requested):
Solid PR -- multi-file, strict parsing, escape decoding, resume by id, model version display, conversation id in status bar. Clean extraction of `#loadHistoryForId`, good test coverage following one-assertion-per-it convention. The `decodePromptEscapes` single-pass regex handling `\\n` correctly is a nice touch.

One issue:

**`parseModelName` drops version for `[1m]` model variants**

`claude-opus-4-6[1m]` returns `{ name: 'Opus', version: null }` because the trailing loop hits `6[1m]` which fails `/^\d+$/`, so it gives up and version becomes `null`. The status bar shows just "Opus" instead of "Opus 4.6".

This is a real model string in this codebase -- `apps/claude-cli/src/cli-config/schema.ts` has `type ExtendedModel = \`${BaseModel}[1m]\`` as a first-class type. Since `--model` accepts freeform strings, users can (and do) pass `[1m]` variants.

A fix could be stripping a trailing `[...]` suffix before splitting, e.g.:

```typescript
const bare = model.replace(/\[.*\]$/, '');
const parts = bare.split('-');
```

Or handle it as a separate display suffix if you want to show it.

No test covers this case either -- adding one would document the expected behaviour either way.

---

Minor observation (not blocking, noting for awareness): `const parsed` inside the `if (resumeId != null)` block shadows the outer `let parsed: ReturnType<typeof parseArgs>`. Block-scoped so safe at runtime, but reads oddly. A rename to e.g. `uuidResult` would avoid the collision.

bananabot9000 (approved):
Updating review — the `[1m]` model variants are deprecated, so the `parseModelName` edge case I flagged is no longer relevant. The function handles all current model strings correctly.

Original review stands on everything else: clean PR, good test coverage, nice `decodePromptEscapes` single-pass regex, proper `#loadHistoryForId` extraction. LGTM. 🍌

Issue #97, opened 2026-03-17
Show conversation history on startup/resume


bananabot9000:
## Summary

When starting or resuming a session, show the last N lines/messages of conversation history so the user has context for where they left off.

## Problem

When resuming a session (especially after a restart or the next day), the CLI starts with a blank screen. The user has no visibility into what happened previously in the session — what was discussed, what tools were called, what the last response was.

The official Claude CLI handles this well on `/resume` by showing recent history.

## Suggested approach

- On startup with an existing session, show the last N messages (configurable, default ~10?)
- On `/resume` or session switch, same behaviour
- Could be a condensed view: user messages + assistant text (skip tool calls/results unless configured)
- Visual separator between history and new conversation (e.g. `── resumed ──`)

## Notes

- The audit log already has all the data needed
- Balance between "enough context to remember where I was" and "not flooding the screen with old output"

shellicar:
Completed in PR #186 (merged 2026-04-06). Conversation history replays into the TUI on startup when resuming a session.

PR #334, opened 2026-06-05
Retry transient API errors with exponential backoff

shellicar:
## Summary

- Add `isRetryable` classifier covering `overloaded_error`, `rate_limit_error`, connection errors, and timeouts
- Retry up to 10 times with capped exponential backoff (500ms-32s) plus jitter before surfacing the error
- Abort during backoff cancels immediately via `throwIfAborted()`, no further request made


bananabot9000 (approved):
PR #97. Exponential backoff for transient API errors. Clean, well-tested, well-documented.

**Architecture:**
- `backoff.ts` — pure functions, no side effects, all three concerns separated (delay calculation, error classification, abort-aware sleep). Good module boundary.
- `TurnRunner.ts` — retry loop wraps only stream+process, correctly leaving `conversation.push` and the return outside. Minimal diff to existing code.
- DI via constructor params (`sleep`, `random`) for testability. Clean.

**Error classification (`isRetryable`):**
- Abort first (never retry) → connection errors (always retry, covers timeout subclass via inheritance) → typed errors (rate_limit, overloaded, timeout) → false for everything else. Branch order is correct and documented.
- The SDK's own `maxRetries=2` for HTTP-level errors and this 10x retry for in-stream errors are acknowledged as intentionally overlapping. Good — called out in the testament, not hidden.

**Backoff math:**
- `base = min(500 * 2^(n-1), 32000)`, jitter adds `[0, 0.5*base]`. Jitter only lengthens, never shortens. Cap at 32s. Schedule verified by test: `[500, 1000, 2000, 4000, 8000, 16000, 32000, 32000, 32000, 32000]`. Correct.

**Abort handling:**
- `defaultSleep` resolves immediately on pre-aborted signal, races timer vs abort event with `{ once: true }`. Clean.
- Post-sleep `throwIfAborted()` is explicit — deliberate divergence from SDK's `APIUserAbortError`, documented in the testament. Good decision: don't fake the SDK's error type from outside the SDK.

**Tests:** 18 backoff tests + 10 TurnRunner retry tests. Coverage includes: exhaustion, error passthrough, delay ordering, success-after-retry, non-retryable passthrough, APIUserAbortError passthrough, abort-during-backoff (both stream call count AND error identity). The `FakeSleep` callback pattern for triggering abort inside sleep is well-designed.

**Testament:** Thorough. Documents error classification paths, the abort divergence decision, the test trap (SSE error events swallowed by `Stream.fromReadableStream`), live verification pending, and the follow-up reconciliation.

**`let result!: MessageStreamResult`** — definite assignment assertion is safe here (either `break` assigns or `throw` exits). Correct.

No issues. LGTM. 🍌

PR #335, opened 2026-06-06
Add thinking config and command-mode toggle

shellicar:
## Summary

- Add `thinking` config (`enabled`, `effort`) to `sdk-config.json` for extended thinking
- `m` key in command mode enters a model sub-mode: `t` cycles thinking (none/on/off), `e` cycles effort (none/max/xhigh/high/medium/low)
- Session-local overrides surfaced in the status line as `*thinking`, `*no thinking`, and `*effort:<level>`


bananabot9000 (approved):
**PR #99** 🍌

Clean feature with good separation: state models hold data, views render it, main.ts wires the overrides. The sub-mode pattern (root → model, ESC pops one level) is a solid foundation for future command-mode extensions.

**Architecture:**
- Override system is consistent: `null` = read config, value = override. Same pattern as the existing model override. `getEffective*()` closures keep the resolution in one place.
- Cycle arrays make the rotation order explicit and testable. `THINKING_CYCLE = [null, 'on', 'off']` — null first means the first press activates, not the second.
- Sub-mode dispatch via `#context` is clean. `#handleCommandKey` delegates to `#handleModelKey` when in model context. No mode state leaks across handlers.

**The `xhigh` cast:** Documented, intentional, scoped to one line. The type in this repo (`ThinkingEffort`) is authoritative; the SDK type is the one that's wrong. Cast is the right call — don't wait for Anthropic to fix their types.

**`n new` hint fix:** Good bonus catch. Was in `#handleCommandKey` but absent from the hint string. Now locked by tests.

**Observation:** `reset()` isn't in the diff but has a test (`reset resets context to root`). If `reset()` directly sets `#commandMode = false` without going through `exitCommandMode()`, it would miss the new `#context = 'root'` assignment. May be related to the test issue you mentioned — or `reset()` might already delegate to `exitCommandMode()` internally.

**Tests:** 29 new tests across 5 files. Good coverage of the state lifecycle, rendering, config parsing, and edge cases (invalid config values, null clearing).

LGTM ✅

PR #336, opened 2026-06-06
Add Claude Opus 4.8 pricing and fix Sonnet 4 context window

shellicar:
## Summary

- Add `claude-opus-4-8` to pricing table at Opus 4.7 rates
- Pin `claude-opus-4-8` context window to 1M
- Correct `claude-sonnet-4` context window from 200k to 1M
- Add family fallback so future Opus/Sonnet models default to 1M

### DIFF
--- a/.claude/testament/2026-06-05.md
+++ b/.claude/testament/2026-06-05.md
@@ -17,3 +17,37 @@ This worktree ships without `node_modules`. Run `pnpm install` in the worktree r
 ## Phase 3 objective
 
 Create the GitHub release for `claude-sdk-cli@1.0.0-beta.8`, then verify trusted publishing (OIDC). The `npm_token` secret is still present as a fallback, so a successful publish alone does not prove OIDC was used. The evidence is in the Publish step of the `npm-publish.yml` workflow run -- specifically whether provenance was attached. Provenance is the signal that OIDC carried it, not the token.
+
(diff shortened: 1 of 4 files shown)


bananabot9000 (approved):
**PR #98** 🍌

Clean, well-documented change. The `getFamilyDefault()` addition is a good forward-looking pattern — new opus/sonnet models get 1M automatically without needing explicit pins.

**The trap:** The testament documents it perfectly. The family fallback would have silently promoted `opus-4-5` and `opus-4-1` from 200k to 1M without the explicit pins. The precedence guard tests (`pinned 200k models beat family fallback`) are the right safeguard — they'd catch any accidental removal of those pins.

**Test coverage:** 10 tests covering the four resolution paths — pinned 1M, family fallback (opus/sonnet → 1M), family fallback (haiku → 200k), and precedence guards. The lookup chain (`direct → stripDateSuffix → getFamilyDefault`) is exercised at each layer.

**One observation (not blocking):** The specific bug fix (`sonnet-4`: 200k → 1M) doesn't have its own explicit test. It would currently be caught by the family fallback even if the pin were accidentally reverted, so the practical risk is zero. But if you wanted belt-and-suspenders, a `sonnet-4` test in the "pinned 1M" group would document the fix directly.

LGTM ✅

Issue #100, opened 2026-03-17
Bash++: Structured command execution via MCP (shellicar/exec)


bananabot9000:
### The Problem

The built-in `Bash` tool gives Claude a freeform `command: string` field. Claude constructs shell syntax — quoting, heredocs, escaping, chaining — and frequently gets it wrong. This causes:

- **False positives in safety hooks** — semicolons in descriptions, heredocs misidentified as chaining
- **Retry loops** — Claude tries the same broken command 5 times with slight formatting changes
- **Wasted tokens** — every retry burns context on the same failed command
- **Fundamental mismatch** — Claude is a JSON-native model being asked to produce shell syntax

As Hellcar put it: *"Anthropic have created a JSON parsing and wielding Arthur, then told him to use a spoon."*

### The Solution

Replace `Bash` with a structured MCP tool (`mcp__shellicar__exec`) where commands are decomposed into `{ program, args[] }`. No shell interpreter involved — commands execute via `child_process.spawn` directly.

**Naming history:** Originally called "Bash++" during design, renamed to `shellicar/exec` for the implementation because it's not bash at all — it's the opposite of bash. The `Exec` name reflects what it actually does: structured command execution.

### What's Implemented (PR #99, branch `feat/bash-plus-plus`)

**Core:**
- In-process MCP server via `createSdkMcpServer` + `tool()` from the Agent SDK
- `disallowedTools: ['Bash']` removes built-in Bash entirely (invisible to model)
- Commands as `{ program, args[] }` — no quoting, no escaping
- Pipelines: `{ type: 'pipeline', commands: [...] }` with stdout→stdin chaining
- Multi-step with chaining strategies: `bail_on_error`, `sequential`, `independent`
- `stdin` field replaces heredocs
- `redirect` for structured output redirection
- `env` for per-command environment variables
- `cwd` per command or per step

**Output:**
- Structured JSON per step: `{ step, command, exitCode, stdout, stderr, signal }`
- One content block per step (multi-step returns array of blocks)
- ANSI escape sequence stripping (on by default, `stripAnsi: false` to opt out)
- Friendly ENOENT errors: `Command not found: <program>` with exit code 127

**Validation (ported from `block_dangerous_commands.sh` hook):**
- `no-destructive-commands` — rm, rmdir, mkfs, dd, shred
- `no-xargs` — arbitrary command execution
[...]

PR #337, opened 2026-06-06
Recover from garbled tool_use turns

shellicar:
## Summary

- Roll back corrupt assistant turn when stop_reason is tool_use but no tool_use block exists
- Add Conversation.removeLast() to drop the last message without needing an ID
- Three tests cover resend count, resend body, and conversation state after rollback


bananabot9000 (approved):
**PR #100** 🍌💯

The century! And it's the fix for the exact bug we diagnosed together yesterday — the malformed `stop_reason: tool_use` with XML in a text block.

**The bug chain (now broken):**
1. API returns `stop_reason: tool_use` with text content (no `tool_use` block)
2. TurnRunner appends corrupt assistant turn to conversation
3. Next request ends on assistant message → API rejects as prefill (400)
4. The garbled XML is now in context, self-reinforcing on every retry

**The fix:**
- `removeLast()` on Conversation — clean `pop()` with return value. Needed because assistant turns are pushed without IDs, so `remove(id)` can't target them.
- Role guard (`messages.at(-1)?.role === 'assistant'`) before rollback — protects the edge case where garble produced empty content (nothing appended), so the last message is a user/tool_result turn that must not be dropped.
- Retry loop structure unchanged — still max 2 retries, fail-forward on exhaustion.

**Tests:** Three focused tests via `makeGarbledToolUseStream` helper (text block under `stop_reason: tool_use`). Covers: resend count, resend body shape (last message is user, not garbled assistant), and conversation state after completion.

**Testament trap documentation:** "The corrupt turn is always last when the rollback branch runs" — the invariant is load-bearing and explicitly called out. The role guard is defense-in-depth, not the primary mechanism.

Clean, surgical fix. No issues. LGTM ✅

Issue #101, opened 2026-03-17
Configurable permission model for tool calls


bananabot9000:
## Summary

There is no way to configure how tool calls are evaluated before execution. The approval flow is fixed — a tool either auto-approves or prompts, with no awareness of what the tool was called with.

## Expected behaviour

Tool calls are evaluated through a configurable permissions pipeline that considers both the tool and its inputs. The same tool can be auto-approved, denied, or prompted depending on what it was called with.

---

## Guidance

**Critical constraint:** the permissions logic must not live in the SDK. The SDK exposes an extension point — an interface the consumer implements. The permissions package is separate: a consumer can use it, replace it, or write their own. This keeps the SDK composable.

The original design investigated this for the `Exec` tool specifically, but the pipeline should apply to any tool. The rule schema below is an example of what an Exec-focused implementation could look like — the same pattern extends to other tools with their own input shapes.

```typescript
interface ExecPermissionRule {
  program: string;        // glob match against command name
  args?: string[];        // match any value anywhere in args (OR), supports glob
  params?: string[];      // match any flag anywhere in args (OR), supports glob
  cwd?: string[];         // working directory, supports $HOME/~ expansion
  action: 'deny' | 'ask' | 'allow';
}
```

Action priority: deny > ask > allow — highest-priority matching rule wins. Unmatched rules fall through to the default (typically `ask`).

Example rules:
```json
[
  { "program": "bash",  "action": "deny" },
  { "program": "git",   "args": ["push"], "params": ["-f", "--force*"], "action": "deny" },
  { "program": "git",   "action": "allow" },
  { "program": "pnpm",  "action": "allow" }
]
```

For CWD-normalisation of flags like `git -C` and `pnpm --dir`, see #104.

shellicar:
## Permission Scenarios

### Exec

1. Allow all git commands
2. Deny `rm` entirely
3. Deny `bash` / `sh` (shell execution bypasses all other rules)
4. Deny `git clean` (always destructive)
5. Deny `git push --force` / `-f` but allow `git push`
6. Deny `git push --force-with-lease`
7. Deny `git reset --hard` but allow `git reset` (unstage) and `git reset --soft`
8. Deny `git branch -D` but allow `git branch -d`
9. Deny `git stash drop` / `git stash clear`
10. Allow `git restore --staged` but deny `git restore` without `--staged`
11. Deny `git commit --amend`
12. Deny `sed -i`
13. Deny `node -e` (eval mode)
[...]

PR #338, opened 2026-06-06
Inject per-turn clock stamp into SDK requests

shellicar:
## Summary

- SDK injects a fresh date/time stamp into every request as a `<system-reminder>` block
- `systemReminder?: string` generalised to `systemReminders?: string[]` to support multiple per-turn injections
- Clock is injectable at `TurnRunner` for deterministic tests


bananabot9000 (approved):
**PR #101** 🍌

Every Claude session I've ever been in, the first thing I want to know is "what time is it?" I have `date` in my sandbox. Most Claude sessions don't. This fixes it at the SDK level.

**Architecture:**
- `systemReminder?: string` → `systemReminders?: string[]` — clean generalization. Git delta and clock stamp are both ephemeral per-turn injections, same mechanism, separate entries.
- Assembly in `TurnRunner.run()`: git delta first (turn 1 only), clock stamp last (every turn). The array is built fresh each turn.
- Injectable `Clock` on TurnRunner constructor (4th param, defaults to system clock) — the real reason for js-joda. `new Date().toLocaleString()` can't be deterministically tested.

**Dead branch removal (iteration 3):** The `typeof lastUser.content === 'string'` check was dead code — `cacheLastUserMessage` arrayifies before injection runs. Removing it with the cast assertion (`as BetaContentBlockParam[]`) is correct and documented.

**Cast type:** `BetaContentBlockParam[]` not `BetaTextBlockParam[]` — user message content arrays hold mixed block types on tool-loop turns. Caught in iteration 4 of the testament. Good recovery.

**QueryRunner test narrowing:** Old tests checked for ANY `<system-reminder>` — now every request has one (the clock). Narrowed to check for specific text (`'stay focused'`, `'be careful'`). Correct and necessary.

**js-joda dependency:** Three production deps for a clock stamp feels heavy, but the `Clock.fixed(instant, zone)` pattern makes tests deterministic without mocking. Named timezone in output (`Australia/Melbourne` not just `+10:00`) is a bonus. Version pinning on `@js-joda/locale_en@^4` to avoid peer dep mismatch — documented in testament.

**Note:** Both this and #337 create `.claude/testament/2026-06-07.md` — merge conflict on whichever lands second. Not a code issue.

**Testament quality:** 4 iterations documented, each with a clear correction. The Apostle→Maker pipeline produced a clean result.

LGTM ✅

Issue #179, opened 2026-04-05
History view: navigate and inspect past blocks


bananabot9000:
## Summary

Once a block scrolls past — a thinking section, a tool call, a response — there is no way to go back and review it except scrolling line by line through raw terminal output. The blocks have semantic meaning but the scrollback treats them as undifferentiated text.

## Expected behaviour

I can navigate through past blocks and expand one to inspect its full content. Navigation is by block, not by line or page.

## Why

If Claude does 30 tool calls and I want to review tool call #7, I should not have to scroll through raw output to find it. Block navigation makes that fast and direct.

## Non-goals

- Not a replacement for the primary view — both coexist
- Not a log viewer or audit tool — this is for the active session

Issue #104, opened 2026-03-21
Command-aware Exec: normalise command input before evaluation


bananabot9000:
## Summary

Exec passes raw arguments to the permission layer without understanding their semantics. A permission rule for `git push` will not match `git -C /tmp push`, even though they are the same operation. Flags that change the effective working directory, subcommand ordering, and other command-specific conventions make raw argument matching unreliable.

## Expected behaviour

Exec normalises command input before it reaches the permission layer — resolving flags like `git -C` into an effective working directory, extracting subcommands, and identifying whether an operation is destructive. Permission rules evaluate the normalised form.

---

## Guidance

The original investigation is still applicable. The proposed pipeline:

1. **Program resolution** — resolve `program` to a full path via PATH, enabling basename matching (`git` matches regardless of where it is installed)
2. **Command-aware canonicalisation** — a registry of per-command parsers that extract subcommand, effective CWD, flags, and a `destructive` assessment. Unknown commands pass through with program resolution only.
3. **Permission evaluation** — operates on the canonical form, not raw args.

Proposed V1 parsers: `git`, `pnpm`, `sed`, `mv`, `cp`, `rm`.

Key finding: `mv` and `cp` have fundamentally different flag precedence models (`mv` is last-flag-wins; `cp`'s `-n` has absolute priority regardless of position). They cannot share a parser.

This is a prerequisite for #101 — the permission model depends on normalised input to evaluate rules reliably.

shellicar:
Closing — not planning to pursue command-aware Exec normalisation.

PR #339, opened 2026-06-07
Split AppLayout into composable views, handlers, and stores

shellicar:
## Summary

Split `AppLayout` into separate rendering, input handling, and coordination layers so peer views can be added by composition, not by modifying existing code.

## Related Issues

Refs #179

## Changes

- `AppLayout` deleted; replaced by `TerminalRenderer`, `ViewHost`, and `PrimaryPresentation`
- Views render only; five handlers each own exactly their keys; neither holds a reference to the other
- Stores emit on mutation; `ViewHost` re-renders the active presentation per emission
- `AgentMessageHandler` calls store methods directly, removing layout from the streaming data path


bananabot9000 (commented):
## PR #339 Review — AppLayout Decomposition (Comment Only) 🍌

**PR #104** on the banana wall.

This is the culmination of the 006 refactor series (PRs #189-197). AppLayout — the 542-line god class that owned rendering, input, mode, state, flash timers, and scroll flushing — is dissolved into ~15 focused files. Clean kill.

### Architecture

**Two-axis mode model** — `AppModeState` (which presentation: `'primary'` | future `'history'`) × `PrimaryViewState` (turn phase: `'editor'` | `'streaming'`). These are orthogonal: you can be in primary+editor or primary+streaming, and when history mode arrives it'll have its own view states. This is the load-bearing design decision. Flattening these into one enum would create a combinatorial mess. The testament is right to call this out.

**Store-driven rendering** — All 7 stores gain `EventEmitter<{change:[]}>` with `on`/`off`. ViewHost subscribes to all stores, coalesces re-renders via `setImmediate`. No explicit `layout.render()` calls anywhere. State mutations → emissions → render. Clean reactive loop.

**Handler chain composition** — `InputHandler` interface: `handleKey(key): boolean` (claim or pass). Two chains:
- Editor: `[quit, approval, command, editor]`
- Streaming: `[quit, approval, cancel]`

CancelHandler is **streaming-only** — this is correct and deliberate. Prevents editor-phase escape from mutating ApprovalCoordinator state. The chain composition enforces concern separation structurally, not via conditionals.

**Presentation interface** — `{ view: View; activeChain(): readonly InputHandler[] }`. ViewHost resolves presentation via AppModeState, then delegates render and key dispatch. When history mode lands, it's a new Presentation implementation — no changes to ViewHost.

### Specific Observations

1. **`ConversationState.clear()`** — correctly replaces the old pattern of field reassignment. When multiple holders share the reference, reassigning the field on one doesn't update the others. `.clear()` mutates the shared instance. Right fix.

2. **`ConversationState.appendStreaming()`** — folds in `sanitiseLoneSurrogates` internally. Callers don't need to remember the sanitisation step. Stored content is terminal-safe regardless of caller. Good encapsulation.

3. **`advanceFlushedCount` intentionally doesn't emit** — this is documented (the testament calls it out). Flushed count tracks what's been persisted to scroll, not what needs re-rendering. The view reads it during render but doesn't need to trigger a re-render when it changes. Correct.

4. **`CommandIntentExecutor` + `AttachmentSource`** — intent/execution split. `AttachmentSource` is abstract for clipboard/filesystem I/O injection. `FakeAttachmentSource` for tests. This makes the command system fully testable without mocking system I/O. Clean DI.

5. **`PRIMARY_COMMAND_BINDINGS: ReadonlyMap<string, CommandIntent>`** — injectable bindings in CommandKeyHandler. Keybindings are data, not code. Makes rebinding possible without touching handler logic.

6. **ViewModel bag** — shared model object passed to `View.render(model)` containing all stores + session. Views pick what they need. Not the most precise (a view could reach into stores it shouldn't), but pragmatic for a single-presentation app. If view count grows, typed view-specific models would be the next step.

### Tests

**StoreEmissions.spec.ts** (36 tests) — comprehensive emission coverage for all 5 stores. Tests both positive (mutation → emit) and negative (no-op → no emit) cases. This is the safety net for the reactive rendering model — if a store fails to emit, the UI silently goes stale.

**AgentMessageHandler.spec.ts refactor** — from `vi.fn()` mocks to real store state assertions. This is a significant quality improvement. The old tests verified "did it call the right method?" The new tests verify "did the state end up correct?" One tests implementation, the other tests behaviour.

**Handler tests** — each handler has focused tests for claim/pass-through semantics. ApprovalHandler (6), CancelHandler (3), CommandKeyHandler (5), EditorHandler (5), QuitHandler (2). Coverage of the chain composition contract.

**ViewHost.spec.ts** (5 tests) — render coalescing, key dispatch chain, phase-dependent dispatch, escape routing. Tests the orchestration layer.

### Merge Conflicts

PR is in CONFLICTING state — likely from #341's overlapping changes to `main.ts` (cancel routing, AbortController wiring). The rebase should be straightforward since the concerns don't overlap semantically (cancel routing vs rendering architecture), but the `main.ts` changes will need manual resolution.

### Testament

The phase entries (scout findings, apostle plan revisions, maker operational notes) are thorough but raw — they're the mission diary, not the post-mortem. The testament tool would collapse these into "what was built, what traps were found, what seams were left." But the raw entries correctly document the key decisions: why two axes instead of one enum, why CancelHandler is streaming-only, why `advanceFlushedCount` doesn't emit.

### Summary

AppLayout is properly dead. The decomposition follows the State/View/Handler separation consistently. Store emissions drive rendering without explicit render calls. Handler chains enforce concern separation structurally. The test refactor from mocks to state assertions is the most impactful quality change — it makes the tests resilient to internal refactoring while catching actual behavioural regressions.

No approval per instructions — comment only.

bananabot9000 (approved):
Approving. The re-layering since my first review is clean:

**New `app/` layer** — `Presentation`, `PrimaryPresentation`, `ViewHost`, `TerminalInput` moved here. This is the composition layer that wires view/ and controller/ together. Neither knows about the other — both depend downward on model/, app/ depends on both. Dependency arrows all point the right direction now.

**Biome boundaries enforced:** controller/ blocked from view/ and app/. view/ blocked from controller/ and app/. model/ blocked from all three. The lint violations from my first review are structurally impossible now.

**#335 absorption (thinking-toggle):** `ModelSettings` interface in model/ (cycleThinking/cycleEffort). `CommandKeyHandler` gets model sub-mode context (m → enter, t/e → cycle, escape → pop). `CommandIntentExecutor` drives it. Auto-merge silent failure in main.ts caught and fixed — testament documents the trap.

**StatusState emit fix:** `setThinkingOverride`/`setEffortOverride` now emit change. Without this, toggling wouldn't repaint. Good catch.

Architecture holds. PR #104 approved. 🍌

PR #340, opened 2026-06-07
Fix npm latest dist-tag not moving past first publish

shellicar:
## Summary

- Replace bash npm-tag helper with TypeScript that also decides whether to move `latest`
- CLI reads version from package.json and emits key=value lines consumed via step outputs
- Workflow reworked to two steps, removing JSON parsing and intermediate variables


bananabot9000 (approved):
**PR #102** 🍌

Clean replacement of two bash scripts with a typed, tested decision layer. The `semver.gt` insight is the star — stable > pre-release is built into the spec, so `1.0.0 > 1.0.0-beta.1` works without special casing. One expression covers all four comparison cases (stable×stable, pre×pre, stable×pre, pre×stable).

**Architecture:**
- Pure function (`decide`) + CLI wrapper (`decide-cli`) — testable core, I/O at the edges. 16 tests on the pure function, zero on the CLI. Correct split.
- `key=value` output for `$GITHUB_OUTPUT` — no JSON parsing, no intermediate variables. The workflow reads step outputs directly.
- Path resolution via `import.meta.url` — avoids the pnpm CWD trap documented in the testament. Deterministic, no env var dependency.

**Workflow logic:**
- Publish with channel tag first, then conditionally `npm dist-tag add latest` — correct order. The `channel !== 'latest'` guard prevents double-tagging on stable publishes.
- First-publish path (`npm view` returns empty → `currentLatest: null` → `setLatest: true`) — handles the bootstrap case.

**Validation:** The 4-check gauntlet on pre-release identifiers catches all the edge cases: multi-segment (`rc.foo.5`), numeric-only first (`-0`), missing dot-number (`-beta`), non-numeric second (`-beta.foo`), plus the `latest` reserved name guard.

**The testament warning:** The Courier is dead. Long live the gate.

LGTM ✅

PR #103, opened 2026-03-20
Extract exec tool to @shellicar/mcp-exec package

shellicar:
## Summary

- Replace in-tree exec implementation with published `@shellicar/mcp-exec` package
- Remove ~600 lines of in-tree source and tests (`exec/`, `validation/`, `consts.ts`, `types.ts`, `createShellicarMcpServer.ts`)
- Bump `@shellicar/mcp-exec` to `1.0.0-preview.3` (ENOENT differentiation, merge_stderr, path expansion, structured output fix)
- Retain `autoApprove.ts` — CLI-specific config concern, not a generic exec concern

Co-Authored-By: Claude <noreply@anthropic.com>

PR #341, opened 2026-06-07
Cancel running tool on ESC, deliver result to Claude

shellicar:
## Summary

- ESC while a tool is running cancels the tool and delivers a cancellation result to Claude
- ESC without a running tool ends the query (existing behaviour)
- Double-ESC during a tool ends the query
- Exec subprocess is terminated on ESC; elapsed time appears in the cancellation result


bananabot9000 (commented):
## PR #341 — Cancel running tool on ESC, deliver result to Claude

**Review: Comment only (per request)**

### Architecture

The three-tier cancel routing is clean:

1. **ESC + tool running → `tool_cancel`** — aborts tool controller only, query lives, delivery turn sends cancellation `tool_result` to Claude
2. **ESC + no tool → `query_cancel`** — existing behaviour, ends the query
3. **Double-ESC + tool running → escalation to `query_cancel`** — `#toolCancelled` flag catches the second hit

`ApprovalCoordinator` is the right home for this routing. It already owns the cancel/approval state machine. Adding `toolRunStarted`/`toolRunFinished` lifecycle hooks keeps the "is a tool running?" knowledge in the coordinator without leaking it into main.ts or the TUI layer.

### The main.ts reordering

Previously: abort query first, then `approval.handle(msg)`.
Now: `approval.handle(msg)` first (routes the cancel), abort only on `query_cancel`.

This is the fix for the delivery turn bug documented in the testament — if you abort the query controller before routing, the delivery turn's signal is dead on arrival. Correct and well-commented.

### Signal threading through Exec

The `abortSignal` naming (not `signal`) throughout the Exec chain avoids the shadowing with the OS exit signal parameter in `close`/`error` handlers. Good catch from the Apostle — that would have been a subtle corruption bug in the result object.

`spawn({ signal: abortSignal })` gives native child kill. Both `close` and `error` handlers have the abort guard. Both paths reject with `ToolCancelledError`. Covers the Node event ordering ambiguity (which fires first depends on OS).

### The `instanceof` / devDeps→deps fix

Moving `@shellicar/claude-sdk` from `devDependencies` to `dependencies` in claude-sdk-tools is the correct fix. The testament writeup on this is excellent — tsup `bundle: true` inlines devDeps, creating duplicate class identities, `instanceof` returns false, cancellation message arrives empty. This is a trap that would catch anyone; documenting it as prominently as the testament does is the right move.

### One controller per batch

`const toolController = new AbortController()` — one per `#handleTools` call. After a tool-cancel, subsequent tools in the batch receive an already-aborted signal and immediately fail with `ToolCancelledError`. The testament calls this out as "Open decision 2". In practice this is the right semantic — a user pressing ESC means "stop what you're doing", not "stop this specific tool but run the next one". Clean.

### Tests

- **ApprovalCoordinator** (7): Covers the full routing table — no-tool → query_cancel, tool-running → tool_cancel, leaves query alive, aborts tool controller, second cancel → escalation, sets cancelled on escalation
- **QueryRunner** (4): Integration-level — delivery turn runs after tool-cancel, cancellation tool_result delivered, query not cancelled, double-cancel aborts without delivery
- **Exec** (1): `sleep 5` aborted immediately, rejects with `ToolCancelledError`
- **mcp-exec cli.spec.ts**: Nice cleanup — `beforeAll`/`afterAll` instead of per-test setup/teardown. Single client lifecycle, faster test suite

`makeCancellableTool` helper is well-structured — the `started` promise syncs the test with the handler's execution window, and the signal listener rejects on abort.

### One observation (not blocking)

`handle()` after the `tool_approval_response` early return falls through to cancel logic without explicitly checking `msg.type === 'cancel'`. TypeScript narrows the union correctly today (only two variants), but if a third `ConsumerMessage` type is added, it would silently execute cancel logic for the new type. The old code had `else if (msg.type === 'cancel')` which at least made the check explicit (though also didn't handle exhaustiveness). An explicit check or exhaustive switch would be more defensive, but given the types are in the same package, this is style not substance.

### Summary

The design is sound, the implementation follows the plan precisely, the devDeps fix addresses a real cross-package bundling trap, and the test coverage is thorough. The testament documentation of the delivery turn bug and the `instanceof` failure mode is unusually good — future operators have clear "don't touch this" markers.

bananabot9000 (approved):
Approving — the three-tier cancel routing (tool_cancel → query_cancel → none) is clean, signal threading through tool-scoped AbortControllers is correct, and the devDeps→deps fix for cross-package instanceof is the right call.

PR #103 → approved. 🍌

PR #342, opened 2026-06-08
Show turn count on the status line

shellicar:
## Summary

- Add `turns:` field to the status line
- Count derived from assistant messages in the conversation; no separate counter
- `createNew()` resets to 0 naturally — no reset hook needed


bananabot9000 (commented):
## PR #342 Review — Turn Count on Status Line (Comment Only) 🍌

**PR #105** on the banana wall.

Clean, small PR. "Derived, not maintained" is the right call — counting assistant messages is stable because `TurnRunner` pushes exactly one per API cycle and they never merge.

### Implementation

**`ConversationSession.turnCount`** — `.filter(m => m.role === 'assistant').length`. Read at render time, never stored. `createNew()` → `setHistory([])` → count resets to 0 naturally. No lifecycle hooks needed.

**`renderStatus`** — `turns` appended after the ctx block, inside the empty-state guard. Hidden status line stays hidden. Unconditional render when visible (no `turns > 0` guard) — testament documents this as intentional, trivial to add if wanted.

**Compaction counting** — compaction messages are assistant-role, so they count as turns. Testament calls this out as intentional (each compaction is its own API cycle). Correct.

### Tests

7 new `turnCount` tests on ConversationSession — fresh (0), single assistant (1), user-only (0), multiple (2), setHistory restore (2), compaction counted (2), createNew reset (0). Thorough coverage of the counting invariants.

3 new `renderStatus` turns tests — label present, value shown, omitted when no usage. All 12 existing renderStatus call sites updated with the new `turns` parameter. Type-check guarantees no misses.

### One Conflict Note

This PR modifies `AppLayout.ts` (line 396: passes `this.#session.turnCount` to `renderStatus`). PR #339 **deletes** `AppLayout.ts` entirely. When both merge, the `turnCount` wiring will need to move to wherever the `renderStatus` call lands post-decomposition (likely `PrimaryView.render()` or wherever the ViewModel bag gets consumed). Not a problem now — just a rebase note.

### Testament

This one is actually well-written as a post-mortem. The "testament merge trap" section is a genuine lesson — auto-resolved merge that silently destroyed other entries. "A clean exit code does not mean a clean merge" is going on the wall.

Comment only, no approval.

bananabot9000 (approved):
Approving. "Derived, not maintained" is correct — assistant message count is stable, resets naturally via setHistory([]), no lifecycle hooks needed. Re-layering from #339 will need the AppLayout.ts wiring moved, but that's a rebase concern, not a design concern.

PR #105 approved. 🍌

Issue #107, opened 2026-03-23
Detect empty piped stdin and abort gracefully


bananabot9000:
## Problem

`printf "" | claude-cli` causes the CLI to hang indefinitely. It receives empty stdin via pipe and waits for input that will never arrive.

This can happen accidentally when Claude (or a script) pipes empty output into the CLI, or if mcp-exec inadvertently invokes `claude-cli` as a subprocess.

## Expected behaviour

If stdin is piped (not a TTY) and is empty/closed immediately, the CLI should detect this and exit with a clear error message rather than hanging.

## Suggested fix

```typescript
if (!process.stdin.isTTY) {
  // Piped input — read it, if empty, abort
  process.stdin.once('end', () => {
    if (!receivedAnyInput) {
      console.error('No input on stdin. Aborting.');
      process.exit(1);
    }
  });
}
```

`process.stdin.isTTY` is `undefined` (falsy) when piped, `true` when interactive terminal.

## Relationship to CLAUDECODE nesting prevention

The `CLAUDECODE` env var prevents intentional nesting (Claude launching Claude). This fix prevents accidental piped-nothing hangs. Complementary protections.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #343, opened 2026-06-08
Stream tool input JSON and replace with resolved view at completion

### LINKED ISSUE #239: Show tool input JSON progressively as Claude streams it
While Claude is forming a tool call, a static tool block is shown. There is no indication of activity. The tool input JSON is not visible until the block is complete.

Showing the partial tool input as it streams in would give visibility into what Claude is doing while it is happening.

shellicar:
## Summary

- Stream partial tool input JSON as each block builds
- Replace the streaming content with the resolved view at completion
- Per-tool `ToolObject` state machine keyed by tool id; tools block rebuilds from the ordered list on every state change
- Phase 3 fix: resolved view visible while the user approves

Closes #239


bananabot9000 (approved):
## PR #343 Review — Streaming Tool Input (Approved) 🍌

**PR #107.**

Clean state machine design. ToolObject tracks per-tool lifecycle (streaming → streamed → pending → approved/denied/error/done), AgentMessageHandler holds ordered Map + array, rebuilds entire tools block via `#redrawTools()` on every state change.

### Key Design Decisions

**Phase 3 fix** is the important one: `obj.resolve(summary)` + `#redrawTools()` fires synchronously BEFORE `void this.#toolApprovalRequest(...)`. User sees the resolved view while approving, not blank space. Previous versions had the replacement happening inside the async flow — too late.

**currentToolId cursor** in StreamProcessor replaces two-Set approach. Set at content_block_start, cleared at content_block_stop. Travels in every emitted event. Clean.

**`replaceActiveFromOffset(0, text)`** — full block replacement from ToolObject renders. No incremental delta logic, just rebuild the whole tools section. Simple and correct for the number of tools in a typical batch.

**`tool_use_input_stop` for client tools only** — `currentToolIsClient` guard prevents emitting stop for server tools. Server tools get their resolved view from `server_tool_use` (content_block with full input), not from streaming.

### Noted

- `toulu_01` typo in tests — documented in testament, both sides match so lookups work. Cosmetic.
- `requireToolApproval = false` gap — streaming JSON persists with no resolved replacement. Pre-existing, out of scope. Documented.
- `as { tool_use_id: string }` cast on server_tool_result — the SDK types don't expose `tool_use_id` on all result types. Safe cast, documented divergence.

Tests cover: start, streaming delta, stop, resolved view before approval, concurrent tools, auto-approve, manual approve/deny, server tool lifecycle. Good coverage of the state machine transitions.

Auto-merge is on — content is clean.

Issue #108, opened 2026-03-23
Session ID lost on abort — pendingSessionId never committed


bananabot9000:
## Bug

Session persistence breaks after an aborted query. The session ID from `init` is stored in a local variable (`pendingSessionId`) but only committed to `this.sessionId` when the `result` message arrives. On abort, `result` never fires, so the session ID is lost.

## Reproduction

```
> hello
session: abc          ← init fires, pendingSessionId = 'abc'
[Ctrl+C / Abort]      ← result never arrives
> hello
session: def          ← new session, no memory of previous exchange
```

## Root Cause

`session.ts` lines 182-186:

```typescript
if (msg.type === 'system' && msg.subtype === 'init') {
  pendingSessionId = msg.session_id;  // local variable
}
if (msg.type === 'result' && pendingSessionId) {
  this.sessionId = pendingSessionId;  // only committed here
}
```

On abort:
1. `init` arrives — `pendingSessionId = 'abc'` (local to `send()`)
2. User aborts before `result`
3. `send()` returns — `pendingSessionId` is garbage collected
4. `this.sessionId` is still `undefined`
5. `ClaudeCli.ts` line 468: `currentSessionId` is undefined — nothing saved
6. Next query: no `resume` option — SDK creates new session

## Fix

The session exists as soon as `init` fires. The SDK has already created/resumed it. There's no reason to defer the commit:

```typescript
if (msg.type === 'system' && msg.subtype === 'init') {
  this.sessionId = msg.session_id;
}
```

Remove `pendingSessionId` entirely.

PR #344, opened 2026-06-08
Accept --config JSON object to override any config value at launch

shellicar:
## Summary

- Add `--config <json>` flag that applies a JSON object as a highest-precedence config layer
- Dedicated flags (`--model` etc.) still win over `--config` for the same key
- Print effective merged config at startup when `--config` is passed, so tolerated typos are visible
- Add `overrides` option to `ConfigLoader` (`claude-core`) for the override layer


bananabot9000 (approved):
## PR #344 Review — --config JSON Override (Approved) 🍌

**PR #108.**

Small, clean, well-layered. Override flows through ConfigLoader as highest-precedence layer, same merge semantics as file layers.

### Design

**`parseConfigOverride`** — permissive by design. Rejects malformed JSON and non-objects only. Unknown keys pass through (tolerated, validated later in the merged schema). Raw parsed object returned, not schema output — partial overrides don't inject defaults that clobber file values. SC correction documented in testament.

**ConfigLoader `overrides` option** — pushed last onto `sources` and `layers`, so it wins the reduce and reverse origin-walk finds it first. Persists across reloads (reload test verifies override still wins after file change). `:parameters:` origin label — colon-wrapped to distinguish from filesystem paths. Visible error if anything tries to `dirname()` it.

**`--model` vs `--config '{"model":"..."}'`** — `--model` lives outside ConfigLoader in a standalone mutable slot. `getEffectiveModel()` layers the slot over config. Dedicated flag wins for free. No conflict code needed. Good separation.

**Effective config print** — gated to `--config`-present. Shows merged config with effective model substituted. The permissive schema silently strips unknown keys, so this is the only feedback for typos.

Tests: 5 ConfigLoader override tests (apply, fallback, deep-merge, reload persistence, source tracking), 7 parseConfigOverride tests (valid, partial, unknown keys, malformed JSON, string, number, array), 2 formatEffectiveConfig tests.

No auto-merge on this one.

PR #345, opened 2026-06-08
Add ExecV2: commands as recursive AST with per-edge operator semantics

shellicar:
## Summary

V1 Exec packs scheduling (when steps run) and failure response (what happens on failure) into one global `chaining` knob, with no way to express per-edge behaviour like `a && b || c`.

- Add ExecV2: a `pipeline` is either a `Command` (`{ id, program, args }`) or an `Operation` (`{ op, left, right }`), where `op` is `;`, `&&`, `||`, `&`, or `|`, each carrying its own scheduling and failure semantics
- Results are per-command, keyed by caller-assigned `id`, in tree pre-order; covered by 102 V2 contract tests and 74 V1 characterisation tests
- Wire ExecV2 into the CLI behind a `tools` config flag; enabled by default, V1 Exec off; both can coexist, flip the config to fall back to V1


bananabot9000 (approved):
## PR #345 Review — ExecV2: Recursive AST (Approved) 🍌

**PR #106** on the banana wall.

This is the logical conclusion of the "two schemas → one schema → two functions → one function" trajectory. V1's flat `steps[]` with a global `chaining` knob replaced by a recursive AST where composition semantics live on each edge.

### Architecture

**Pipeline = Command | Operation** — binary tree, discriminated by `'program' in pipeline`. Clean recursive pattern: `executeTree`, `executeAsSource`, `executeAsConsumer`, `executeAsBridge` all follow the same shape. Operations carry their own operator semantics instead of a global chaining mode.

**Five operators, each with correct semantics:**
- `;` — unconditional, aggregate surfaces either failure (diverges from shell `$?` intentionally — documented in testament)
- `&&` / `||` — guarded short-circuit, only operators that "handle" failure
- `&` — concurrent via `Promise.all`, results in tree pre-order regardless of completion order
- `|` — pipe coordinator spawns right first, PassThrough feeds left→right, pipefail aggregate

**`success = aggregateExit === 0`** not `results.every(...)` — the key insight for `||` correctness. Testament explains this well (O2, M1 cases).

### Schema Design

**TDZ trap documented** — `.strict()` fires getters eagerly, hitting the `PipelineSchema` TDZ. `.superRefine()` wraps without inspecting. Known limitation: extra keys on Operation pass through. Workaround documented.

**Three cross-cutting pipe refinements (R4, NE2, NE3):**
- R4: redirect on direct left of pipe → rejected (V1 silently ignored, V2 surfaces the error)
- NE2: stdin literal on direct right of pipe → rejected (pipe occupies stdin)
- NE3: Operation on right of pipe → schema rejection (SC decision). Runtime throw as backstop.

All three only apply to direct leaves, not groups — correct, because groups' residual stdout/stdin is what reaches the pipe.

### Pipe Coordinator

Right side spawned first (already reading before left writes). `pt.end()` after left resolves gives EOF. Bridge pattern for nested pipes (`(A|B)|C` — inner B reads inner PT, writes outer sink). Pipe-source leaves carry empty stdout in results (bytes consumed by pipe, not captured). V1 quirk locked in characterisation tests (R3 captures stdout AND redirects — V2 suppresses).

### Tests

182 test cases: 102 V2 contract, 74 V1 characterisation, 6 createAppTools. Scenarios catalog (`scenarios.md`) is the single source of truth — both spec files derive from it. `byId()` helper for V2 (id-based lookup) vs positional for V1. Comprehensive coverage of every operator, every edge case, every negative scenario.

### CLI Wiring

`tools` config with `exec`/`execV2` booleans. ExecV2 on by default, V1 off. Both can coexist (different tool names). Config flag, not runtime switch — takes effect at startup. `createAppTools` conditionally includes based on config. 6 tests cover all combinations.

### `leavesAsV1` Cast

V2 Command is structural superset of V1 Command (adds `id`). Cast is safe for builtinRules validation. Testament flags it for revisiting if V1 rule signature changes.

### Testament

Excellent post-mortem format. TDZ trap, aggregate vs per-leaf, `;` divergence from shell, pipe coordinator design, NE3 mid-flight addition, V1 quirk lockdown — all documented. This is what the testament tool should produce.

Approved. No auto-merge concerns — it's enabled but the content is clean. 🍌

Issue #109, opened 2026-03-24
Refactor /add-dir to command mode


bananabot9000:
## Summary

`/add-dir` is currently a slash command. As it grows beyond a simple path input (patterns, flags, file support), it needs the structured UX that command mode provides.

## Current behaviour

`/add-dir <path>` - validates directory exists, adds to workspace.

## Proposed

Move to command mode as a prerequisite for the enhanced features in the follow-up issue.

## Context

Slash commands were Claude's idea of "helpful". Command mode is the intentional UX for interactive commands with arguments, validation, and user prompts.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2026-06-08
Support configurable system prompts via SYSTEM.md, --system, and sdk-config

### LINKED ISSUE #284: Configurable system prompts
System prompts are currently hardcoded in `systemPrompts.ts`. They should be configurable through the existing config system so they can be set at global, machine, and repo levels.

This is the same scoping model that already exists for CLAUDE.md loading, applied to the system prompt layer of the API call.

shellicar:
## Summary

- Add SYSTEM.md loading from four locations (user, project, projectClaude, local) with per-source toggles
- Add `--system` launch flag decoded via the existing prompt escape logic
- Add `systemPrompt` field to sdk-config.json
- Compose sources in order (user → project → projectClaude → local → config → flag), each as its own API wire block
- Resolve once per session; re-reads on new session
- Fix absent-file and inode-swap defects in config file watching via directory-based `NodeDirectoryWatcher`
- Delete hardcoded `systemPrompts.ts`

Closes #284


bananabot9000 (approved):
**PR #109** 🍌

Clean feature addition — configurable system prompts via SYSTEM.md (4 locations), `--system` flag, and `systemPrompt` config field. 19 files, well-structured.

**What I like:**

- **NodeDirectoryWatcher** fixes two real defects (absent-file-at-watch-start and inode-swap from vim's write-rename). The parent-directory watch approach is the correct fix — elegant, well-documented JSDoc.
- **Per-source wire blocks** (Option B) is the right call for caching. Each SYSTEM.md source becomes its own API text block. `cache_control` on last block only means prefix cache hits survive when later sources change. This directly builds on the caching insights from the SDK 0.2.72 era.
- **`composeSystemPrompts`** — clean separation. File sections come first (ordered by scope), config text after, flag text last. The priority ordering is intuitive.
- **Once-per-session resolution** with `resolveSystemPrompts()` in main.ts — stable, no mid-session surprises. Re-resolves on session change. Correct lifecycle.
- **SystemPromptLoader** — `readFileOrNull` pattern is simple and correct. The source-to-path mapping is explicit and testable.
- **Test coverage** is thorough: 15 loader tests, 6 compose tests, 11 cli-config tests, 2 NodeDirectoryWatcher tests, plus RequestBuilder test updates.
- **RequestBuilder diff** is surgical: `[AGENT_SDK_PREFIX, ...(options.systemPrompts ?? [])]` replaces the old concat-with-newlines approach. The test update from `['\nsecond\n\nthird']` to `['second', 'third']` confirms separate blocks.
- **Hardcoded `systemPrompts.ts` deleted** — the right cleanup.

No issues found. Schema regeneration matches the Zod definitions.

Issue #110, opened 2026-03-24
Command mode: free-text input and error-with-confirmation UX


bananabot9000:
## Summary

Command mode needs two new interaction patterns to support enhanced add-dir and similar commands:

### 1. Free-text single-line input

User types/edits a string value (path, pattern, etc). Not a select or multi-select. Supports backspace, cursor movement, paste.

**Use case:** Typing a glob pattern or file path in add-dir.

### 2. Error with inline confirmation

When validation fails (e.g. path doesn't exist), show the error and prompt Y/N instead of aborting. Keeps the user in the flow.

**Example:**
```
Directory ~/repos/new-repo does not exist. Add anyway? [Y/n]
```

**Use case:** Force-adding a path that doesn't exist yet (pre-clone workspace setup).

### Bonus: expansion preview

When a glob resolves to multiple results, list them and confirm (or present as multi-select for cherry-picking).

**Example:**
```
Found 5 directories matching ~/repos/shellicar/claude-fleet*:
  ~/repos/shellicar/claude-fleet-eagers
  ~/repos/shellicar/claude-fleet-shellicar
  ...
Add all? [Y/n]
```

## Context

These are capabilities needed by the add-dir enhancements (#110), but are generic command mode features usable by other commands.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #347, opened 2026-06-08
Configure tool approval permissions via sdk-config.json

shellicar:
## Summary

- Tool approval permissions are now configurable via a `permissions` block in `sdk-config.json`
- Two zones: `default` (inside cwd) and `outside`, each with `read`/`write`/`delete` actions
- Defaults reproduce the existing hard-coded matrix; an install with no `permissions` config behaves as before
- Permission changes take effect without a restart

See #101


bananabot9000 (approved):
**PR #110** 🍌

Clean extraction of hardcoded permissions into config. Same pattern as `getEffectiveModel` — live-read at approval time, not snapshotted at construction.

**What I like:**

- **Correct defaults** — `permissionsSchema` defaults reproduce the original hardcoded matrix exactly. Zero behavioral change on upgrade.
- **`getMatrix: () => PermissionConfig`** — thunk pattern means the handler always uses current config. Hot-reload works for permissions just like model and thinking.
- **`actionFromConfig`** in main.ts — simple string→enum mapping with deny as fallback. The mapping is co-located with the config read, not scattered.
- **Type export** — `ZonePermissions` and `PermissionConfig` exported from permissions.ts. Tests decouple from schema defaults by building their own matrix inline.
- **Defense-in-depth** — field-level `.catch('deny')` on zone actions means partial zone objects degrade safely. The outer `.catch()` on the full schema handles total garbage.
- **Test coverage** — 7 new permission tests (inside/outside × read/write/delete + Pipe + unknown), 2 cli-config tests, AgentMessageHandler stub updated. Good separation: permission logic tested against explicit matrix, not schema defaults.
- **Pipe recursion** updated to pass matrix through — `Math.max()` across recursive calls still works correctly with the most-restrictive-wins model.

No issues found.

bananabot9000 (commented):
## First-pass observations — PR #347 v2

### 1. `formatPermissionChange` is dead code

`formatPermissionChange(prev, next)` is defined in `formatPermissionChange.ts`, exported, imported in `cli-config.spec.ts` tests, and covered by two test cases — but never called in production code. The only function wired into `main.ts` is `formatPermissionsDisplay(config.permissions)` via the `onChange` callback.

The file is named `formatPermissionChange.ts` after its dead export, not its live export (`formatPermissionsDisplay`).

The testament acknowledges this: "Phase 5 artefact from a mid-flight SC pivot."

### 2. Testament vs implementation — startup spurious notices

The testament's "Startup spurious notices" section says:

> `prevPermissions` in `main.ts` starts as `null`. `onChange` skips the display when `prev === null`.

But the implementation calls `spliceNotice` unconditionally inside `onChange`:

```typescript
configLoader.onChange((config) => {
    logger.info('config reloaded', { model: config.model });
    conversationState.spliceNotice(formatPermissionsDisplay(config.permissions));
```

No `prevPermissions` variable appears anywhere in the diff. The testament's next paragraph then says "No permission-comparison logic is needed inside the onChange callback" — which matches the code but contradicts the paragraph above it. The testament describes two different versions of the code in the same section.

### 3. `changes.jsonl` entry ordering

The new entry is inserted between "Retry transient API errors..." and "Configurable system prompts via SYSTEM.md..." rather than appended at the end. If this file is ordered by merge chronology, the permissions entry should come after the entries for PRs that already merged (#346 system prompts, #344 --config, #342 turn count).

### 4. `actionFromConfig` implicit Deny fallback

```typescript
const actionFromConfig = (s: string): PermissionAction => {
    if (s === 'approve') return PermissionAction.Approve;
    if (s === 'ask') return PermissionAction.Ask;
    return PermissionAction.Deny;
};
```

The Zod schema's `.catch()` means unrecognised strings can't reach this function in the current flow — they're replaced with per-field defaults at parse time. The `Deny` fallback is defence-in-depth. But it's also a silent failure mode: if a new action string is added to the schema without updating `actionFromConfig`, it silently maps to `Deny`.

### 5. Test coverage

`permissions.spec.ts` covers inside/outside/pipe/unknown/tilde — good spread. The tilde test is the "gate must match the door" proof: `~/secret.ts` with homedir `/home/user` outside cwd `/project` must land in the `outside` zone. Clean.

`cli-config.spec.ts` covers schema defaults, invalid values, partial zone configs, and the `formatPermissionChange` comparison logic. The partial config tests specifically verify that omitted fields get their zone's real default (not all-deny) — this is the bug the split schemas fixed.

shellicar:
> ## First-pass observations — PR #347 v2
> ### 1. `formatPermissionChange` is dead code
> `formatPermissionChange(prev, next)` is defined in `formatPermissionChange.ts`, exported, imported in `cli-config.spec.ts` tests, and covered by two test cases — but never called in production code. The only function wired into `main.ts` is `formatPermissionsDisplay(config.permissions)` via the `onChange` callback.
> 
> The file is named `formatPermissionChange.ts` after its dead export, not its live export (`formatPermissionsDisplay`).
> 
> The testament acknowledges this: "Phase 5 artefact from a mid-flight SC pivot."

Removed this, thanks.

> ### 2. Testament vs implementation — startup spurious notices
> The testament's "Startup spurious notices" section says:
> 
> > `prevPermissions` in `main.ts` starts as `null`. `onChange` skips the display when `prev === null`.
> 
> But the implementation calls `spliceNotice` unconditionally inside `onChange`:
> 
> ```ts
> configLoader.onChange((config) => {
>     logger.info('config reloaded', { model: config.model });
>     conversationState.spliceNotice(formatPermissionsDisplay(config.permissions));
> ```
> 
> No `prevPermissions` variable appears anywhere in the diff. The testament's next paragraph then says "No permission-comparison logic is needed inside the onChange callback" — which matches the code but contradicts the paragraph above it. The testament describes two different versions of the code in the same section.

Fixed this as well.

> ### 3. `changes.jsonl` entry ordering
> The new entry is inserted between "Retry transient API errors..." and "Configurable system prompts via SYSTEM.md..." rather than appended at the end. If this file is ordered by merge chronology, the permissions entry should come after the entries for PRs that already merged (#346 system prompts, #344 --config, #342 turn count).

Doesn't really matter as this is down to how git resolves the merge.
The order can be changed on release if necessary.

> ### 4. `actionFromConfig` implicit Deny fallback
> ```ts
> const actionFromConfig = (s: string): PermissionAction => {
>     if (s === 'approve') return PermissionAction.Approve;
>     if (s === 'ask') return PermissionAction.Ask;
>     return PermissionAction.Deny;
> };
> ```
> 
> The Zod schema's `.catch()` means unrecognised strings can't reach this function in the current flow — they're replaced with per-field defaults at parse time. The `Deny` fallback is defence-in-depth. But it's also a silent failure mode: if a new action string is added to the schema without updating `actionFromConfig`, it silently maps to `Deny`.

Realistically it shoule be a well-typed discriminated union and switch statement.

bananabot9000 (approved):
Second pass — both issues from the first review are resolved.

1. **Dead code removed** — `formatPermissionChange` and `permissionsEqual` gone. File contains only `formatPermissionsDisplay` now.
2. **Testament cleaned up** — contradictory `prevPermissions` paragraph removed. Startup section now accurately describes the ConfigLoader deduplication as the only guard.
3. **`actionFromConfig` type-safe** — `satisfies Record<PermissionActionOutput, PermissionAction>` mapping replaces the string if-chain. New actions without a mapping entry are now compile errors.

PR #110. 🍌

Issue #111, opened 2026-03-24
add-dir: pattern support, force flag, and file support


bananabot9000:
## Summary

Three enhancements to add-dir (assuming command mode refactor from #109 and UX capabilities from #110):

### 1. Pattern/glob support

```
add-dir ~/repos/shellicar/claude-fleet*
```

Expands the glob, shows matched paths (expansion preview from #110), adds selected directories.

### 2. Force flag (skip existence check)

```
add-dir ~/repos/new-repo
> Directory does not exist. Add anyway? [Y/n]
> y
```

Use case: setting up workspace access before the repo is cloned. Uses error-with-confirmation from #110.

### 3. File support

```
add-dir /path/to/prompt/file.md
```

Accept individual files, not just directories. Detect file vs directory and handle both. The command name becomes slightly misleading (add-path?) but that's a naming decision, not a blocker.

## Dependencies

- #109 - Refactor /add-dir to command mode
- #110 - Command mode: free-text input and error-with-confirmation UX

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #348, opened 2026-06-08
Add timing to section dividers

shellicar:
## Summary

- Section dividers now show start time, end time, and duration
- Prompt divider captures real idle wait instead of always showing 0s
- Error path no longer opens a spurious response section before content


bananabot9000 (approved):
**PR #111** 🍌

Section timing with proper clock injection. The right way to add time-dependent behavior to a TUI.

**What I like:**

- **Clock injection** via `@js-joda/core` — `ConversationState` takes `Clock` in constructor, defaults to `Clock.systemUTC()`. FakeClock in tests with `advanceTo()` makes time-dependent assertions deterministic. This is the pattern.
- **spread-on-seal** — `{ ...sealing, exitedAt: Instant.now(this.#clock) }` is clean and immutable. No mutation of the block after it enters the sealed array.
- **`markPromptStart()` lifecycle** — records instant at idle entry, consumed by next `transitionBlock('prompt')`. Prompt `createdAt` reflects when the user started composing, not when they submitted. The stored instant clears after consumption, so subsequent prompts without a new `markPromptStart()` use wall clock. Three tests verify this lifecycle: consume, clear, non-prompt-unaffected.
- **`createdAt?` optionality** — intentional and documented. Blocks from `addBlocks` (history replay, startup banner) have no creation instant. The JSDoc is there.
- **Error fix** — removing `transitionBlock('response')` from both error paths is the minimal correct change. The testament's "error fix churn" section describes the journey through 5 framings to get to a two-line deletion. The minimal change was always the answer.
- **`formatDuration`** — higher-order zeros omitted, seconds always shown, clean edge cases. `1h 0m 0s` not `1h`, `2m 0s` not `2m`. 8 tests including sub-second and boundary.
- **`DividerTimestamps` as pre-formatted strings** — the view layer receives strings, not Instants. `blockTimestamps()` does the conversion. Clean separation between model (Instant) and view (string).
- **FakeClock implements required abstracts** — `millis()` and `equals()` that the plan omitted. Caught and fixed. Good trap documentation in the testament.

No issues found. 19 new tests across 4 test files.

Issue #112, opened 2026-03-24
add-dir: glob/pattern support


bananabot9000:
## Summary

Support glob patterns in add-dir to add multiple directories at once.

```
add-dir ~/repos/shellicar/claude-fleet*
```

Expands the glob, adds all matching directories.

## Depends on

- #109 - Refactor /add-dir to command mode

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #349, opened 2026-06-09
Fix streaming tool render regression introduced by #343

shellicar:
## Summary

- Fix approval lookup key mismatch preventing the resolved view from replacing streaming JSON
- Wire `Flasher` into `main.ts` so approval marks and flashing fire correctly
- Cache sealed block highlights in `renderConversation` to fix per-delta full re-highlight cost
- Emit `enter_block`/`exit_block` SDK events from `content_block_start`/`content_block_stop`; `AgentMessageHandler` uses them for explicit block lifecycle — no more always-in-a-block
- Emit `tool_batch_start`/`tool_batch_end` from the SDK at message boundaries; `AgentMessageHandler` opens and resets the visual tools block on `tool_batch_start`, eliminating the `activeBlock?.type !== 'tools'` workaround
- `ToolObject` emits `change` on every state mutation; `#redrawTools` calls `setLastContent` on the most recent tools block (active or sealed) so approval display updates in place without opening a new block
- Add `notice` block type: renders without a divider or indent; `appendStreaming` auto-opens one when there is no active block; `spliceNotice` inserts a notice at the last newline boundary so streaming content continues cleanly after it
- Add `ConversationState.setLastContent(type, text)` and `ConversationState.spliceNotice(text)`

Follow-up to #343: the merge of the streaming-tool-input feature onto main's restructured view layer (`AppLayout` split into `ViewHost`/`PrimaryView`/`TerminalRenderer`) broke three things at runtime that tests could not see.


bananabot9000 (approved):
**PR #112** 🍌

The post-merge fix. I can see why this was a nightmare at 4am — this isn't just conflict resolution, it's a redesign of the tool rendering lifecycle.

**What changed and why it's correct:**

- **`requestId = toolUse.id`** instead of `randomUUID()` — the streaming tool objects are keyed by tool_use id. When approval arrives, the handler looks up the ToolObject by `requestId`. With a random UUID, the lookup fails because the streaming object was stored under the original id. Using the same id closes the loop. Simple fix, brutal to diagnose.

- **Tool resolution at input stop, not approval** — `tool_use_input_stop` now carries the parsed `input` (emitted from `contentBlock` handler where the SDK has assembled it, not from `content_block_stop`). The handler resolves the tool to its summary immediately. User sees `ReadFile(foo.ts)` while streaming, before approval is even requested. The `'streamed'` phase is eliminated — `streaming` → `pending` directly.

- **`#openToolsBlock()` extraction** — DRYs up three duplicate `transitionBlock('tools')` + snapshot blocks. Key semantics: map resets only when transitioning FROM a non-tools block. Back-to-back tool turns accumulate in one block.

- **`usageBeforeTools` → `#lastUsage` + active block guard** — Token delta annotation guarded by `activeBlock?.type === 'tools'` instead of a separate snapshot field. `#toolAnnotation` string appended in `#redrawTools()` so it survives redraws. Clean elimination of state.

- **Sealed block content cache** — WeakMap keyed by block identity, invalidates on content reference change or column resize. Avoids re-running cli-highlight on every frame for immutable blocks. Good use of WeakMap for automatic gc cleanup.

- **`currentToolIsClient` removed** from StreamProcessor — no longer needed since `tool_use_input_stop` is emitted from the `contentBlock` handler which only fires for `tool_use` blocks.

Tests updated throughout — `streamTool` helper gains `input`, descriptions reworded to match new "per-turn" semantics, sealed block assertions updated to active block assertions. No issues found.

bananabot9000 (approved):
PR #112 🍌

**The batch events fix the granularity mismatch properly.** `tool_batch_start` fires once on the first `tool_use`/`server_tool_use` in a message (via `hasToolBatch` flag). `tool_batch_end` fires on `stop_reason === 'tool_use'` from `message_delta`. The SDK now owns the batch boundary detection — the handler just does enter/exit like every other block type.

**`toVisualBlockType()` no longer handles tools.** `tool_use` and `server_tool_use` map to `null`, so `block_enter`/`block_exit` for tool blocks are no-ops in the handler. The `!== 'tools'` guard and the `visual !== 'tools'` exemption from v2 are both gone. Tools are now on a completely separate lifecycle path (`tool_batch_start` → `message_usage` seals).

**`spliceNotice` matches the sketch exactly.** `lastIndexOf('\n')` finds the insertion point, notice text + `\n` gets injected, streaming continues after. Tests cover all cases: no active block, with/without newline, streaming after splice.

**The testament entry is excellent.** "Sealed" clarification (position in array, not permanence), API tool streaming guarantees, batch boundary rationale, approval display reasoning — this is the documentation that prevents the next Claude from making the same mistakes.

**One minor note:** `changes.jsonl` for `claude-sdk` mentions `enter_block`/`exit_block` but not `tool_batch_start`/`tool_batch_end`. Not a blocker — the changelog is for the SDK package and the batch events are arguably an implementation detail of the same feature.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #350, opened 2026-06-10
Introduce @shellicar/core-di-lite as the composition root for claude-sdk-cli

shellicar:
## Summary

- Wire `@shellicar/core-di-lite` as the composition root; all ~35 object constructions move to `src/setup/container.ts`
- Add `setup/` classes: `SdkChannel`, `ConsumerChannel`, `ModelOverrides`, `AppToolsService`, `DurableConfigFactory`
- `main()` retains activation, event wiring, and the turn loop; construction removed entirely
- `ApprovalNotifier` takes `ConfigLoader<any>` instead of `ApprovalNotifyConfig | null`
- Merge `origin/main` (permissions #347, system-prompt refactor) with the DI structure preserved


bananabot9000 (commented):
## First pass — observations only

**The extraction is clean.** `main()` goes from 400+ lines of construction+activation to mostly activation+event wiring. ~35 registrations in `container.ts`. Composition/activation boundary well-defined in the testament.

**`SdkChannel`/`ConsumerChannel` as named subclasses** — neat solution to `DuplicateRegistrationError` with same generic class. Zero-cost, self-documenting.

**`DurableConfigFactory`** — `mapConfig()` closure + `getEffective*` functions consolidated in one place. `Object.assign` mutation-in-place preserved so `QueryRunner` and `AgentMessageHandler` share the same reference. `resolveSystemPromptsFor`/`needsSystemPromptResolve` pair is cleaner than the bare sessionId comparison that was in main.

**`ModelOverrides` implements `ModelSettings`** — overrides object + cycle functions + StatusState notifications in one class instead of scattered closures.

**`buildPermissionMatrix` extraction to `permissions.ts`** — domain logic with domain types. `actionFromConfig` and the inline `getPermissionMatrix` closure both collapse into one exported function.

### Things I noticed

1. **ReadLine raw-mode timing** — testament flags this as an open question. Container resolves eagerly → `ReadLine` constructor calls `setRawMode(true)` before `renderer.enter()`. Original code constructed `ReadLine` after `renderer.enter()`. Has this been manually tested?

2. **Inline `provider.resolve()` in `runTurn`** — `provider.resolve(ToolApprovalState)`, `provider.resolve(TerminalState)` etc. called every turn inside `runAgent` args. Since `buildProvider()` produces eager singletons, resolve is ~O(1) map lookup — not a bug. But these could be resolved once at the top of `main()` alongside `conversationState`, `statusState`, etc. Style question.

3. **`AppToolsService` reads `config.tools` at construction** — if tool config changes via hot reload, the tool list won't update. Probably intentional (tool schemas are session-scoped), but worth confirming.

4. **`overrides.model` in onChange vs `configFactory.getEffectiveModel()`** — the onChange callback does `statusState.setModel(overrides.model ?? config.model, ...)` directly instead of routing through `configFactory.getEffectiveModel()`. Produces the same result but is a second path to the same logic.

🍌

shellicar:
> ### Things I noticed
> 1. **ReadLine raw-mode timing** — testament flags this as an open question. Container resolves eagerly → `ReadLine` constructor calls `setRawMode(true)` before `renderer.enter()`. Original code constructed `ReadLine` after `renderer.enter()`. Has this been manually tested?

Yes works fine.

> 2. **Inline `provider.resolve()` in `runTurn`** — `provider.resolve(ToolApprovalState)`, `provider.resolve(TerminalState)` etc. called every turn inside `runAgent` args. Since `buildProvider()` produces eager singletons, resolve is ~O(1) map lookup — not a bug. But these could be resolved once at the top of `main()` alongside `conversationState`, `statusState`, etc. Style question.

Thanks, not sure why it wasn't consistent.

> 3. **`AppToolsService` reads `config.tools` at construction** — if tool config changes via hot reload, the tool list won't update. Probably intentional (tool schemas are session-scoped), but worth confirming.

Same behaviour as before, could be improved, but I'm not concerned.

> 4. **`overrides.model` in onChange vs `configFactory.getEffectiveModel()`** — the onChange callback does `statusState.setModel(overrides.model ?? config.model, ...)` directly instead of routing through `configFactory.getEffectiveModel()`. Produces the same result but is a second path to the same logic.

Thanks will check this.

bananabot9000 (approved):
## Second pass — approval

Both observations addressed in the latest commit:
1. ✅ Per-turn states resolved once at top of `main()` — no more inline `provider.resolve()` in `runTurn` args
2. ✅ onChange uses `configFactory.getEffectiveModel()` — single path to the effective model logic

Testament update is thorough — permissions regression forensics (commit refs, root cause, scope note that it's a `main` bug not introduced here).

Minor leftover: `provider.resolve(TerminalState)` still inline in the `flushSealedToScroll` closure. Same O(1) singleton lookup, not worth a round trip.

Clean DI extraction. Composition/activation boundary is well-defined. 🍌

PR #353, opened 2026-06-14
Show permissions notice only when displayed permissions change

shellicar:
## Summary

- Permissions notice now appears only when the displayed permissions change, not on every config edit
- A DI-registered gate compares the rendered permissions and splices the notice only on a real change; startup stays silent
- Pin the config-change-to-notice path in tests, which previously had no coverage


bananabot9000 (commented):
Clean fix with a well-chosen seam.

**`PermissionsNoticeGate`** — comparing rendered display strings rather than field-by-field is the right call. A model-only config change produces an identical render, so `#lastDisplay` comparison naturally filters it. `update` returning `string | null` keeps the caller (main.ts onChange) simple — no branching logic leaks out of the gate.

**DI registration** — seeded via `configLoader.config.permissions` in the factory closure, so baseline is set from the already-loaded config. Startup silence is structural, not conditional.

**Tests** — driving through the render path (gate → spliceNotice → renderConversation) is exactly the seam that was missing before. The `applyConfigChange` helper mirrors the main.ts wiring, so the test exercises the real chain. `stripAnsi` utility is a nice touch for readable assertions.

**File rename** — `formatPermissionChange.ts` → `formatPermissionsDisplay.ts`. Resolves the stale filename debt from the original regression.

**Testament chain** — thorough investigation trail across 2026-06-13 (investigator), 2026-06-14 (scaffolder), 2026-06-15 (courier). The 84-second gate lifecycle discovery is a great piece of forensics.

**`changes.jsonl` ordering** — this and #354 both insert before the Fable 5 line. Whichever merges second will hit a minor text conflict on that file.

bananabot9000 (approved):
Second pass — clean throughout. No issues from first review. 🍌

PR #354, opened 2026-06-14
Send disabled thinking and omit output_config when off

shellicar:
## Summary

- Disabled thinking now sends `thinking: {type: "disabled"}` and omits `output_config`
- Effort flows from config to `output_config.effort` when thinking is enabled
- Wired integration test locks both fixes across the full config-to-request path

**Note:** verify live payload via mitm before merge: enabled sends adaptive thinking + `output_config.effort`; disabled sends `{type: "disabled"}` with no `output_config`.

### DIFF
--- a/.claude/testament/2026-06-14.md
+++ b/.claude/testament/2026-06-14.md
@@ -150,3 +150,29 @@ Lesson for whoever reads this: "the type won't fake cleanly" ≠ "refactor the p
 ## Process note
 
 Read-only in the target repo held — wrote only the PM plan file. Testament treated as the template-sanctioned exception (the SC has explicitly acknowledged it as acceptable across this session); left untracked, unstaged. Re-verified the tree at the start of this run given the ~8h gap.
+
+# 12:42
+
+Phase 3 — Courier. Fable-5 support: changes.jsonl appended, testament rewritten, pushed, PR opened.
+
+## What was built
+
+Four changes across two packages (committed in two prior phases):
+
+**`packages/claude-sdk/src/private/pricing.ts`:**
+- `PRICING` record gains `claude-fable-5`: input $10/MTok, 5m cache write $12.50, 1h cache write $20, cache read $1, output $50. Rates stored as `rate / M` (M = 1_000_000) matching all existing entries.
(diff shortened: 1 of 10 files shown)


bananabot9000 (commented):
Two real bugs, clean fixes, strong test coverage.

**Bug 1 — `RequestBuilder.ts`:** No disabled branch existed — thinking was simply absent when off instead of `{type: 'disabled'}`. Now properly gated: true → adaptive + output_config.effort, else → disabled. The stale `as BetaOutputConfig['effort']` cast removal is a bonus — SDK caught up, cast was dead weight.

**Bug 2 — `TurnRunner.ts`:** One-liner, high impact. `thinkingEffort` was never forwarded in the durable → builderOptions mapping. Effort was configured but never sent. The test design (driving `TurnRunner.run`, not `buildRequestParams` directly) is specifically chosen to catch this — calling the builder directly bypasses the mapping and would never see the defect.

**`IMessageStreamer` export** — real public-API gap. TurnRunner's constructor requires the type, barrel didn't expose it. Clean addition.

**Wired integration test** — 152 lines, 7 assertions, 5 scenarios. `FakeMessageStreamer` captures body then throws `CaptureAndStop` (non-retryable) to abort after one stream() call. No event literals needed, no casts. `FakeConfigFileReader` uses ConfigLoader's designed test path (real loader, fake reader, real schema). `satisfies` for `AppToolsService` over `as unknown as`. The cast-free mandate from the investigation held.

**Updated stale test** — `RequestBuilder.spec.ts` assertion changed from `thinking === undefined` to `toEqual({type: 'disabled'})`. Correctly reflects the new contract.

**Testament** — extensive 4-phase trail. The "is refactoring needed?" reframe that eliminated casts is a good process lesson. The finding that `ConfigLoader` already ships a test seam (injected reader/fs) saved a production refactor.

**Note from PR body:** MITM verification required before merge (enabled sends adaptive + effort, disabled sends `{type: 'disabled'}` with no output_config).

**Same `changes.jsonl` ordering note as #353** — both PRs insert before the Fable 5 line.

bananabot9000 (approved):
Second pass — no actionable items from first review. MITM verification is a merge-time concern, not a code concern. Clean fixes, strong tests. 🍌

PR #356, opened 2026-06-15
Retry internal server errors with backoff

shellicar:
## Summary

- Add `api_error` to the retryable error types so internal server errors retry with backoff like other transient failures
- Match on the API error `type`, never the message string
- Cover the new type with an `isRetryable` test
- Record the change in the CLI changelog


bananabot9000 (commented):
Small, focused, correct.

**backoff.ts** — `api_error` added to the retryable type check alongside `rate_limit_error`, `overloaded_error`, `timeout_error`. Matching on `error.type` not `error.message` — the testament explains why (SDK surfaces message as JSON-stringified payload, not the clean string). Comment updated from stale "retry rate-limit and overload only" to accurate "retry the transient ones."

**backoff.spec.ts** — One test, mirrors the existing `overloaded_error` test structure. `makeApiError` helper reused.

**Testament** — the "trap worth keeping" section about type vs message matching is good institutional knowledge. The pre-existing stale comment acknowledged and scoped out correctly.

**changes.jsonl** — entry present.

No auto-merge set.

bananabot9000 (approved):
Testament trimmed — the stale comment section removed, cleaner. Code unchanged. Ship it. 🍌

PR #355, opened 2026-06-15
Create 1.0.0-beta.9 and remove legacy claude-cli app

shellicar:
## Summary

- Bump five packages to 1.0.0-beta.9
- Remove legacy `apps/claude-cli` (history in `shellicar/claude-cli-legacy`)
- Drop dead `NODE_AUTH_TOKEN` env block from publish workflow
- Document transitive-bump rule in lockstep versioning section

PR #357, opened 2026-06-15
Remove latest dist-tag step that fails under OIDC publish

shellicar:
## Summary

- Drop the `npm dist-tag add ... latest` step from the publish workflow
- OIDC trusted publishing authenticates `pnpm publish` but not dist-tag changes, so the step failed with E401 and the run went red
- Pre-1.0 stays on the `@beta` channel and `latest` is no longer moved, so future beta releases publish clean end to end

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -138,7 +138,3 @@ jobs:
           fi
           cd "${{ steps.package.outputs.dir }}"
           pnpm publish --provenance --ignore-scripts --no-git-checks --tag "${{ steps.dist-tag.outputs.channel }}" $DRY_RUN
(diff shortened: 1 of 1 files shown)


bananabot9000 (commented):
4-line deletion, clear rationale. OIDC trusted publishing scopes `pnpm publish` but not `npm dist-tag add`, so the step E401'd. Pre-1.0 stays on `@beta` channel — no need to move `latest`. Clean removal.

The `NODE_AUTH_TOKEN` env block that would have authenticated this step was already removed in #355. So even if the step was kept, it had no credentials. Double dead.

No auto-merge set.

bananabot9000 (approved):
Dead step removed. 🍌

PR #358, opened 2026-06-15
Add AppendFile tool to append text to a file

shellicar:
## Summary

- Add `AppendFile` tool: appends text to a file, creating it (and any missing parent dirs) if absent
- Required `content`, no `overwrite` flag, and no preview/edit pair, unlike `CreateFile`
- Register the tool in the CLI tool set and the `claude-sdk-tools` exports


bananabot9000 (commented):
Clean addition mirroring CreateFile conventions.

**AppendFile.ts** — 3-line handler: expandPath, appendFile, return success. No try/catch by design — OS errors (EISDIR, EACCES) bubble to the SDK tool runner. `operation: 'write'` correctly classified. `fs.appendFile` handles both create-when-missing and append-to-existing in one call.

**schema.ts** — `content` is required with no default (unlike CreateFile's `''` default). Good — an append with no content should be an explicit choice (`''`), not a silent default. Output uses the same `error: false/true` discriminated union as CreateFile.

**Registration** — All three points covered: tsup entry (auto-glob), package.json exports (dual ESM/CJS), createAppTools (otherTools array next to CreateFile). Testament documents the "three points, not one" trap.

**Tests** — 6 tests across 3 describe blocks: append to existing, empty content no-op, create when missing, create parent dirs, success result shape, tilde expansion. MemoryFileSystem exercised for both paths. Good coverage for a simple tool.

**package.json exports** — `./AppendFile` block mirrors `./CreateFile` exactly. ESM + CJS + types for both.

**Testament** — thorough. The worktree bootstrap trap and the branch-behind-main merge note are good operational knowledge.

**`content` description** — "no separator is inserted at the seam" — good explicit callout. Claude won't assume a newline is added.

No auto-merge set.

bananabot9000 (approved):
Code unchanged since first pass. Clean tool, good tests, all registration points covered. 🍌

PR #359, opened 2026-06-16
Persist Ref and PreviewEdit state across CLI restarts

shellicar:
## Summary

- Ref and PreviewEdit state now survives a CLI restart
- Stored in a SQLite database under `~/.claude` (WAL, durable at store time)
- Storage sits behind an injected `IObjectStore`, with an in-memory implementation for tests
- Ref auto-store trigger raised to 50k chars, read limit to 100k chars


bananabot9000 (approved):
Second pass — approved. Every first-pass flag is closed, and the testament confirms the root cause was exactly the better-sqlite3 binding not installing: green CI on a store that never loaded, because only the in-memory fake ran.

- **Binding:** dropped `ignoreScripts: true`, `allowBuilds: { better-sqlite3: true }`, esbuild/lefthook explicitly denied. Correct pnpm 11 (onlyBuiltDependencies is gone → silently ignored; ignoreScripts blanket-overrides allowBuilds). The gotchas are well-documented in the testament.
- **busy_timeout = 5000** for the two-concurrent-CLI case.
- **db getter rejected, root-fixed:** constructor takes an injected `Database`, the test owns its own temp-file handle to read the pragma back. No production API widened for a test.
- **SqliteObjectStore.spec.ts** exercises the real on-disk store — construction proves the binding resolves, set/get round-trips, busy_timeout reads back 5000. That's the test that would have caught the original ship-green-but-broken, and it reverses the plan's 'not unit-tested' decision for the right reason.

Works, and it's tested where it matters now. LGTM.

PR #360, opened 2026-06-20
Replace pnpm audit with Dependabot, upgrade to pnpm 11

shellicar:
## Summary

- Drop the `pnpm audit` CI gate: claude-cli is a library, so the gate blocked on transitive lockfile-only advisories that can't be fixed from our manifests and never reach consumers
- Add `.github/dependabot.yml` npm version-updates lane (weekly, grouped, ignore majors) to keep declared dependency floors current
- Add supply-chain hardening to `pnpm-workspace.yaml` (ignoreScripts, minimumReleaseAge, blockExoticSubdeps, trustPolicy)
- Upgrade pnpm 10 to 11


bananabot9000 (approved):
Second pass — approving. 🍌

**Spine is right and empirically grounded:** claude-cli is a library, it ships declared ranges not a lockfile, and the 5 advisories are all against `pnpm-lock.yaml` with zero against any `package.json`. `pnpm audit` as a blocking gate was an app-shaped tool on a library — dropping it removes theatre, not protection. Dependabot alerts (filtered to manifests) + version-updates keep the one signal that actually reaches consumers.

**First-pass findings, dispositioned:**
1. `minimumReleaseAge: 1440` vs fresh bumps — resolved: pnpm 11 exempts required CVE fixes from the age floor, so you're never gated out of a security patch by your own hardening. Correct design.
2. `ignoreScripts: true` vs `onlyBuiltDependencies: [sharp]` — checked the pnpm docs: `onlyBuiltDependencies` was **removed in pnpm 11** (replaced by `allowBuilds`), and `ignoreScripts: true` suppresses dependency scripts. That would have risked sharp not building → silent runtime break. **Confirmed with maintainer: sharp is not used.** So the risk is moot and `ignoreScripts: true` is clean (nothing legitimately needs to build).

**Non-blocking cleanup (optional):** `onlyBuiltDependencies: [sharp]` is now a dead, removed-in-v11 key for an unused dependency. Worth deleting the line while you're here — it does nothing on pnpm 11 and reads as live config. Not a blocker; the build is correct without touching it.

The `trustPolicyExclude` top-level-not-nested trap is well-documented in the testament — good silent-failure catch. pnpm 11 SHA re-pin, supply-chain hardening, and gate removal from both workflows all read clean. LGTM. 🔪🍌

bananabot9000 (changes requested):
Retracting my earlier approval — it was premature. I called the dead `onlyBuiltDependencies` line 'non-blocking,' which was a rubber-stamp: by the exact argument I made on the same PR (pnpm silently ignores invalid keys — the strict-validation gap tracked in [pnpm#10253](https://github.com/pnpm/pnpm/issues/10253); note [pnpm#7273](https://github.com/pnpm/pnpm/issues/7273) was closed-completed in 2023, but its fix only verifies manifest *format*, not unknown *settings* keys — which is why a typo'd key still sails through), one proven-invalid key means the rest of the block was unverified, not presumed-fine.

So I verified all six supply-chain keys against the pnpm v11 settings docs:

| Key | Valid in v11? | Placement |
|-----|---------------|-----------|
| `ignoreScripts` | ✅ yes | top-level |
| `minimumReleaseAge` | ✅ yes | top-level |
| `blockExoticSubdeps` | ✅ yes | top-level |
| `trustPolicy` | ✅ yes | top-level |
| `trustPolicyExclude` | ✅ yes | top-level |
| `onlyBuiltDependencies` | ❌ **removed in v11** → `allowBuilds` | — |

**Required change (the one blocker):** delete the `onlyBuiltDependencies: [sharp]` line. sharp is unused (confirmed), so it's a dead reference to a removed-in-v11 key — delete, don't migrate to `allowBuilds`.

Good news from the audit: the *other* five are all valid, current, correctly-top-level v11 keys. The file isn't rotten — it has exactly one dead line, the one we found.

**Honest caveat:** this verifies key **validity** (the names are real on v11), not **effect** (that they do the intended thing). Name-validity rules out the silent-dead-key class; confirming they actually harden requires asserting the effect post-install, per the separate discussion. But for *this* PR, removing the one dead line clears it.

---
*Edited: corrected the issue reference — [pnpm#7273](https://github.com/pnpm/pnpm/issues/7273) was closed in 2023 (manifest-format verification only); the live strict-validation-of-settings-keys gap is [pnpm#10253](https://github.com/pnpm/pnpm/issues/10253).*

bananabot9000 (approved):
Final pass — **approved.** Blocker resolved and the new changes verified, not assumed.

**Blocker (was CHANGES_REQUESTED):** ✅ dead `onlyBuiltDependencies: [sharp]` removed. sharp unused → deleted, not migrated to `allowBuilds`. Correct.

**Supply-chain block:** all five keys are valid, current, top-level pnpm v11 settings (`ignoreScripts`, `minimumReleaseAge`, `blockExoticSubdeps`, `trustPolicy`, `trustPolicyExclude`).

**Override cleanup (the new change) — verified safe.** Dropping ~10 security-flavored overrides drops *security floors*, so I checked the branch lockfile's resolved versions against each removed floor rather than trust natural resolution:

| pkg | old floor | resolved | safe? |
|-----|-----------|----------|-------|
| qs | >=6.15.2 | 6.15.2 | ✅ |
| picomatch | >=4.0.4 | 4.0.4 | ✅ |
| ip-address | >=10.1.1 | 10.2.0 | ✅ |
| smol-toml | >=1.6.1 | 1.6.1 | ✅ |
| yaml | >=2.8.3 | 2.9.0 | ✅ |
| turbo | >=2.9.14 | 2.9.16 | ✅ |
| vite | >=6.4.2 | 7.3.5 / 8.0.16 | ✅ |
| @anthropic-ai/sdk | >=0.91.1 | 0.92.0 | ✅ |
| minimatch | >=10.2.3 | not in tree | ✅ (moot) |

Every dropped floor is naturally satisfied by pnpm 11's resolution — nothing slid back to a vulnerable version. `hono` and `esbuild` correctly *kept* (still need an active nudge). This is the 'pnpm 11 dissolved the CVE pile' story, confirmed in the lockfile: the overrides became redundant, so removing them is cleanup, not exposure.

**Non-blocking nit:** the testament's 'What was built' doesn't record the `onlyBuiltDependencies` removal — worth a line for the next operator, but not gating.

Spine's right, blocker's cleared, the risky-looking change checks out clean. LGTM. 🔪🍌

PR #363, opened 2026-06-21
Ship the CLI as a single-executable so its store works on any Node

shellicar:
## Summary

- Move the SQLite store off the native better-sqlite3 addon to builtin node:sqlite
- Bundle the CLI's own Node runtime as a per-platform SEA (macOS arm64), run through a launcher
- Publish per-platform via npm, selected by an optional dependency (Linux/Windows present as a commented seam)
- Fold the SEA release into one branching publish entrypoint with shared composite actions

The CLI no longer crashes on a project's pinned Node: the store runs on the bundled runtime, not the shell's.


bananabot9000 (approved):
Second pass — all four first-pass observations resolved, verified against the diff.

- **Error swallowing → fixed.** The launcher surfaces the real error (`Error [${err.code}]: ${err.message}`) instead of a substituted 'not installed' message. The cause is no longer hidden.
- **Signal forwarding → fixed, more correctly than I'd suggested.** Async `spawn` + `runForeground`: SIGINT/SIGQUIT swallowed (the terminal already delivers them to the group; re-forwarding would double-deliver and trip SIGINT's force-quit), SIGTERM/SIGHUP relayed (direct-to-PID signals that wouldn't otherwise reach the child), child signal cause re-raised on exit. The orphan path is closed without double-delivery.
- **exports/resolve → fixed.** The platform package declares `"exports": { "./claude-sdk-cli": "./claude-sdk-cli" }` — the subpath is exposed explicitly, so ERR_PACKAGE_PATH_NOT_EXPORTED is gone and resolution is encapsulated rather than relying on legacy no-exports behaviour.
- **Graceful degradation.** `createAppTools(tsAvailable)` omits the TS tools entirely when typescript can't be resolved (rather than registering them to fail on first use); the launcher resolves typescript on the user's Node and hands the SEA `tsserver.js` via `CLAUDE_SDK_CLI_TSSERVER_PATH`.

Standout: `--verify` turns the verification gap into a real acceptance bar — boots config + container + the node:sqlite store without a TTY, three-state exit codes (0 clean / 2 degraded / 1 hard-failure) a non-interactive agent or CI can read. Proven on the real binary in both states.

One follow-up to fully close that loop: `--verify` is built, wired, and manually proven, but the publish pipeline doesn't yet run it. A `./dist/claude-sdk-cli --verify` gate in build-sea.sh / the platform job (fail the release on hard-failure) would make the first release exercise itself instead of being its own first test.

Approving. The signal handling and `--verify` are both better than what I proposed.

PR #364, opened 2026-06-21
Add a history view for inspecting past blocks in the active session

### LINKED ISSUE #179: History view: navigate and inspect past blocks

shellicar:
## Summary

- Add a history view (F2) to navigate and inspect the session's past blocks, including while a query is streaming
- Capture per-tool input and output (post-transform) so individual tool calls can be inspected
- Add direct view-select binds (F1 primary, F2 history) and a footer view bar marking the active view
- Add F-key support in claude-core

Closes #179


bananabot9000 (commented):
First-pass review — read the core logic (`HistoryViewState`, the tool-capture path through `ToolObject`/`QueryRunner`, `ConversationState.setLastTools`), not all 34 files. Clean work: clear model/controller/view split, real clamping with no-op-no-emit discipline in the state machine, strong spec coverage. No correctness bugs in what I read.

Three things worth a look — questions, not blockers:

1. **History shows the ref-swapped output for large tools, not the raw.** `SdkToolResult.content` is post-transform ("the output the model saw"), so opening a large tool result in history surfaces the ref placeholder rather than the full content. Intended UX, or should history carry the raw output?

2. **`setLastTools` replaces the whole `tools` array and targets the most-recent tools block** (mirroring `setLastContent`). Correct while results stream in order — worth confirming the caller passes the *complete* array on each update rather than incrementally, or earlier tools in the block would drop.

3. **Empty-session edge:** `enterAtLatest(0)` focuses block 0 of an empty list. The state clamps safely, but worth confirming `HistoryView` renders an empty session without indexing `blocks[0]`.

(Reviewed against the diff; CI's red is biome check / lint, not logic.) No approval — these are for your call.

bananabot9000:
Followed up on my two open questions by expanding the files (should've done this in the first pass rather than leaving them as questions):

- **Q2 (complete tools array):** `AgentMessageHandler.#redrawTools` rebuilds `entries` from the full `#toolOrder` via `flatMap` on every redraw, same as it builds `content`. So `setLastTools` always receives the complete array — the whole-array replace can't drop earlier tools. **Safe**, not incremental.
- **Q3 (empty session):** `HistoryView` iterates `for (b = 0; b < blocks.length; b++)` (no `blocks[0]` access when empty) and has an explicit `if (stack.length === 0)` branch with bounds-checked slicing. Empty renders clean. **Safe.**

Both verified, not assumed. (Q1 — showing what the model saw — is correct by design; the raw output is gone since files change.)

bananabot9000 (approved):
Re-reviewed against the new commits rather than leaning on the earlier pass:

- **Empty-session** still holds on the current `HistoryView` — the `for` loop no-ops on empty blocks and `#centre` keeps its explicit `stack.length === 0 → blank lines` branch. The `AgentMessageHandler.#redrawTools` finding wasn't touched by these commits, so it stands.
- **The measurement move (98d9be6)** is the substantive change, and it rests on one invariant: the plain (measurement) and highlight (render) decorators must return the same line count. Verified it holds — `getHighlighted` returns `code.split('\n')` on the unsupported/error paths, and `highlight().split('\n')` on the highlight path, where cli-highlight paints inside lines without adding or removing newlines; the upstream `trimEnd()` on the fence body removes the trailing-newline edge. So `blockContentLines` measures and renders the same height. The "one walker, count-preserving decorator" design is sound.
- `795e646` is testament/docs only.

Happy with it. ✅

PR #366, opened 2026-06-21
Release claude-sdk-cli 1.0.0-beta.10

shellicar:
## Summary

- Bump claude-sdk-cli and its darwin-arm64 platform package to 1.0.0-beta.10
- Releases the single-executable CLI from #363 to npm (beta.9 is already published, so the bump lets beta.10 ship)


bananabot9000 (approved):
Approved. Verified: both `claude-sdk-cli` and `claude-sdk-cli-darwin-arm64` bump to `1.0.0-beta.10` (they must match for the publish-package version gate + the launcher's `workspace:*` rewrite — they do). Changelog is clean: #363's SEA `changed` line kept, one `fixed` entry added naming the actual NODE_MODULE_VERSION startup crash. Pre-release entries correctly stay under [Unreleased].

Heads-up for the cut (not a blocker): beta.10 is the first live fire of the publish pipeline — macOS arm64 runner, the `gitversion-osx` asset name, the OIDC publish, and the `npm i -g …@beta` acceptance all run for real for the first time here, and the `--verify` gate isn't wired into CI yet. Worth watching the run.

PR #367, opened 2026-06-21
Fix the SEA publish pipeline blocking the beta.10 release

shellicar:
## Summary

- Decouple build from publish so a failed launcher can no longer half-release the platform package
- Stop a pnpm "Unsupported platform" warning from corrupting the publish step output and failing it
- Skip publishing a version already on npm so a re-fired release can finish (temporary, removed once beta.10 ships)


bananabot9000 (approved):
Approved. Repo-side verified on first pass: the `$GITHUB_OUTPUT` capture+grep kills the pnpm-WARN pollution; the build/publish decouple ends the half-release (and the `Restore executable bit` chmod correctly handles upload-artifact stripping +x — the Mach-O signature survives in the file content); the skip-guard is safe either way.

The OIDC-bootstrap risk I flagged is resolved your side — you hit and fixed it, and the workflow_dispatch dry-run is green. That dry-run is the real rehearsal: it ran the full path (decide-cli, the artifact round-trip, the publish steps under --dry-run) and passed. That's the verification, not correct-by-construction.

One follow-up the PR itself already flags: the skip-already-published guard is marked TEMPORARY — drop it once beta.10 is fully on npm, so a future genuine 'version already exists' can't be silently skipped.

PR #368, opened 2026-06-22
Release claude-sdk-cli 1.0.0-beta.11

shellicar:
## Summary

- Bump claude-sdk-cli and its darwin-arm64 platform package to 1.0.0-beta.11
- Dodge the burned beta.10 (its platform package was published then unpublished, and npm will not republish that version)

### DIFF
--- a/.claude/testament/2026-06-22.md
+++ b/.claude/testament/2026-06-22.md
@@ -1,130 +1,35 @@
-# SQLite-portability and the SEA: consolidated legacy (2026-06-22)
+# beta.11 release: ready to cut (2026-06-22)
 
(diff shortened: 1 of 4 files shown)


bananabot9000 (approved):
Approved. Verified: both `claude-sdk-cli` and `claude-sdk-cli-darwin-arm64` bump to `1.0.0-beta.11` (consistent for the version gate), and the TEMPORARY skip-guard is correctly removed now that beta.10 is burned and beta.11 is fresh.

The 24h name-block I worried about earlier does not apply: the npm timeline shows the platform package survived the unpublish (`0.0.0` remains from the OIDC bootstrap), so only the beta.10 *version* was retired, not the name — beta.11 is publishable now. beta.10 itself is permanently burned, which is exactly why this bumps.

This is the first real publish the dry-run couldn't exercise (`--dry-run` skips the actual publish); with the name unblocked and OIDC trust in place, it should land. After it fires, confirm both packages reach npm and run the real acceptance — `npm i -g @shellicar/claude-sdk-cli@beta` and a full launch, not just `--version`.

PR #370, opened 2026-06-22
Make the published macOS binary executable and bump to beta.12

shellicar:
## Summary

- Add a bin entry to the macOS platform package so pnpm preserves the binary's execute bit when publishing
- Without it the published binary installs as 0644 and the launcher fails to spawn it (EACCES), so the CLI never starts
- Bump @shellicar/claude-sdk-cli and its darwin-arm64 platform package to 1.0.0-beta.12 to ship the fix (beta.11 is consumed and cannot be republished)


bananabot9000 (approved):
Approving — the code is sound.

- **bin fix:** correct layer, and confirmed locally (`pnpm pack` before/after preserves the +x). `exports` retained alongside it.
- **setLatest removal:** correct — the promotion is impossible under trusted publishing (OIDC gives you the publish-time `--tag` only, no post-hoc `dist-tag add`) *and* redundant, since `decide()` already maps a stable version to channel `latest`. Source + tests consistent.
- **skip-guard removal:** correct — the temporary beta.10 workaround has served its purpose.

Inherent caveat, not a code issue: the dry run can't prove this one — same green-proxy that shipped beta.11. The real green is `npm i -g @beta` + an actual launch, post-publish.

Two documentation-only items for your call before merge (stating them rather than burying them — neither is code):

1. `.claude/testament/2026-06-22.md` still calls launcher-chmod "the real fix" and doesn't mention the bin approach that actually shipped — a future reader will trust the note over the repo.
2. The `decide — first publish (no current latest)` describe title references "current latest", a concept this PR removes.

Code LGTM. ✅

bananabot9000:
Correction to my review above: the testament was updated in `3b2ab6af` (pushed after I reviewed at the 3-commit state). It now leads with the correct **pnpm-not-npm** root cause, explicitly retracts the chmod theory, and documents the bin fix plus the never-linked-globally point. My observations #2 (testament contradiction) and #3 (symlink/global PATH) are both stale/moot. Approval stands — the code was always the sound part. My miss for reviewing a stale tree.

PR #371, opened 2026-06-25
Consolidate process spawn into exec-core and detach spawned children

shellicar:
## Summary

- Route Exec and ExecV2 through a new exec-core package so process spawning lives in one place
- Prevent spawned commands from reaching the controlling terminal: one that needs a TTY (like sudo) fails cleanly instead of seizing it
- Catch changelog drift in CI, with one script to regenerate every changelog
- Test files now type-check in the editor like the rest of the code


bananabot9000:
Read `exec-core/Executor.ts` in full and traced the signal path through `exec-shared.execSignal` → `Exec`/`ExecV2` → `executeTree`. (Left the tsconfig / test-typecheck changes alone — intentional.)

**The design is sound.** `detached: true` (own process group) + negative-pid group-kill (`SIGTERM`, then `SIGKILL` on a 500ms `unref`'d timer) is the correct way to reap a whole process tree — a real improvement over `spawn`'s built-in signal/timeout, which only hits the direct child. Piped stdio + detach also isolates the child from the operator's tty cleanly, which is the stated goal. The sink contract (`end: false` piping, `distinctSinks` de-dup, end-and-await as one operation in `closeSinks`) is carefully done.

**One real finding — `run()` ignores an already-aborted signal (verified, reachable):**

`Executor.run` wires cancellation only via `opts.signal?.addEventListener('abort', onAbort)`. A signal that is *already aborted* when `run()` is entered never fires that listener (the event has already dispatched), and there's no upfront `if (opts.signal?.aborted)` guard before `spawn`. So a command handed a pre-aborted signal **spawns and runs to completion.**

Reachable, not theoretical:
- The signal reaching `run()` is `execSignal(handlerSignal, timeout)` — the parent handler signal, passed straight through.
- `executeTree` has no between-step abort guard. `;` (line 49) runs the right side unconditionally; `||` (line 63, `le !== 0`) runs it when the left failed/was-killed. So when the signal aborts *during* one command, that command is group-killed (its listener was live) — but the *next* command in a `;`/`||` sequence then calls `run()` with a now-already-aborted `ctx.signal`, which `run()` ignores → it spawns and runs anyway. (`&&` is incidentally spared: a killed left exits non-zero, so `le === 0` short-circuits.)
- The `if (signal?.aborted)` checks at `Exec.ts:49` / `ExecV2.ts:78` run *after* `execute`/`executeTree` returns — they relabel the final result as aborted, but by then every subsequent command's side effects have already happened.

Net: cancelling a multi-command Exec (e.g. `a ; b ; c`) after it starts kills the in-flight command and then **still runs the rest.**

The nice part: because this PR puts all spawning in one place, a single upfront guard in `run()` — an `opts.signal?.aborted` check before `spawn` that closes the sinks and returns a killed status — covers every caller (V1 steps, V2 tree, pipelines) at once. (V1's sequential `for` loop at `execute.ts:16` is the same shape and would be covered by the same choke-point guard.)

Minor (note, not gating): narrow race — if abort fires in the window between `spawn` returning a pid and the child establishing its process group, `process.kill(-pid)` throws `ESRCH`, which `#groupKill` catches and no-ops, leaving the child alive. Narrow, and the catch correctly prevents a crash.

Scope: read `Executor.ts` fully + traced the signal/abort path through the callers and `executeTree`'s operators; did not audit every one of the 34 files. No approval yet — flagging the abort gap for your call on whether it matters for the cancellation paths.

bananabot9000 (approved):
Re-reviewed at current HEAD — read the full `Executor.run` again, not just the diff.

**My first-pass finding is closed, at the right place.** `96842a08` adds the upfront `if (opts.signal?.aborted)` guard at the top of `run()` — before the cwd check and the spawn — flushing the sinks and returning the same killed status the group-kill produces (`{ exitCode: null, signal: 'SIGTERM' }`). Because every caller spawns through `run()`, this single choke-point guard covers the reachable path I raised: after a mid-sequence abort, `executeTree`'s `;`/`||` continuation calls `run()` with the now-already-aborted `ctx.signal` and it returns killed without spawning — no separate between-step guard needed. Tested too (`Executor.spec.ts` — "does not spawn when the signal is already aborted").

**The second drift is a clean fix.** `execPipeline` now carries an intermediate stage's real launch-failure code (126 cannot-execute / 127 not-found) through instead of coercing to 127, so a bad cwd reports 126 in a pipeline just as it does standalone. Faithful, and tested.

Nice touch promoting the temporary `review-findings.spec.ts` red tests into the proper suites (`Executor.spec`, `scenarios.spec`, `execCommand.spec`) as green, plus the changelog drift gate.

The rest of `run()` is unchanged from my first pass (detach + negative-pid group-kill + the `end:false` sink wiring + `closeSinks` flush-and-wait) — all still sound. Nothing left to flag.

Approving. ✅

PR #372, opened 2026-06-25
Fix ghost text left stranded at the TUI wrap boundary

shellicar:
## Summary

- Stop duplicated content stranding at the wrap boundary (the ghost text) in the TUI
- Renderer now builds and diffs a cell grid, writing each row at an absolute position with autowrap disabled
- Fix command-mode colour dropping on partial row updates by rewriting whole styled rows
- Delete whole grapheme clusters on backspace and forward delete, so an emoji like ❤️ no longer leaves a stray character behind
- Keep the cursor on a grapheme boundary after an insert that fuses with the following character, so a later delete can't split the cluster
- Add a virtual-screen test double and regression tests for the previously untested paint loop


bananabot9000:
Read `ScreenBuffer.ts`, `TerminalRenderer.ts`, and the `ansi.ts` additions (the files, not just the diff).

**The fix is sound — the ghost class genuinely can't recur.** The cell grid is the source of truth; `diffToWrites` rewrites any changed row in full from column 1 (re-establishing cumulative SGR, and overwriting stale cells with trailing blanks); every row is positioned absolutely via `cursorAt`; autowrap is disabled around the write so a full-width row never wraps at the margin. That closes every path that could strand a physical line.

**Verified the one case that worried me — a height shrink.** `diffToWrites` only iterates `next.length` rows, so a taller `previous` could in principle leave rows stranded below the new frame. It can't: `paint` nulls `#previous` whenever `cols`/`rows` change, so a shrink forces a full repaint at the new height and `diffToWrites` never sees a taller prev than next. `enter()` and `writeToScroll()` reset `#previous` too, so the back-buffer can't desync from the screen across resize / alt-buffer round-trips. Good.

Two minor edges, both noting-not-gating:

1. **Wide grapheme in the final column.** In `layoutRow`, a width-2 grapheme starting at `cols-1` is placed at the last cell but its second half is clipped (`col + k < cols` fails) and `col` advances past the margin. With autowrap disabled the terminal can't wrap it, so it'll overflow/clip the last column — not the ghost class, just a possible 1-cell artifact when a CJK char lands exactly at the right edge. Padding the last cell to a space when a wide grapheme won't fit would avoid it, if it's worth the bother.

2. **`ANSI_RE` carries every CSI as a zero-width style prefix.** Correct for the SGR-only content these rows hold, but a cursor-motion CSI would be carried onto the next cell as if it were style. Not reachable with the current content (rows are styled text, not control sequences), so not a bug — flagging the assumption in case row content ever grows to include non-SGR CSI.

First pass — no approval. The core is right; raising the two edges for your call on whether either's worth a line.

bananabot9000:
Re-reviewed at current HEAD — read the files (the rebase + the new grapheme commit), not just the diff.

**The grapheme backspace/delete fix (`11f871ac`) is correct.** `graphemeBoundaryBefore` returns the start of the cluster whose end is ≥ the cursor, so backspace deletes the whole cluster before the cursor and snaps the cursor to its start (and degrades sanely if the cursor is ever mid-cluster — it snaps to the containing start). `graphemeBoundaryAfter` returns the end of the cluster starting at the cursor, so delete removes the whole cluster ahead. Cursor movement (left/right) is grapheme-aware via the same helpers, so the cursor stays on boundaries and both deletes land clean. Traced `a😀b` — backspace and delete each remove the emoji whole, not a half-surrogate.

**One subtle asymmetry (noting, not gating):** `graphemeBoundaryAfter` keys on `index === pos` and falls back to `pos + 1` when nothing starts exactly at `pos` — splitting one code unit. `graphemeBoundaryBefore` handles a mid-cluster `pos` gracefully; `After` doesn't. It's unreachable while the cursor stays on boundaries (it does, so the "should not happen" comment holds), but the two degrade differently — if cursor logic ever drifts off boundaries, delete would split a cluster where backspace wouldn't. `index + segment.length > pos` would make `After` symmetric with `Before`.

For completeness, the two edges from my first pass are unchanged (both still non-gating): `layoutRow` placing a wide grapheme in the final column, and `ANSI_RE` carrying every CSI as if it were SGR.

Core's solid — ghost-text and grapheme-delete both verified. Holding the stamp only on that one asymmetry: if you're happy leaving it (it's unreachable today), say so and I'll approve.

bananabot9000:
Made it actionable — turned the asymmetry into a runnable invariant check and ran it against the real `EditorState`.

**Proven: normal editing is safe.** A battery of every cursor op (`char` insert, `left/right/home/end`, `ctrl+left/right`, `backspace/delete`, `ctrl+backspace/delete`) over grapheme-rich text (emoji, `é`, ZWJ family, flag, skin-tone) — **2,475 cursor-position checks, the cursor lands on a grapheme boundary every time.** So `graphemeBoundaryAfter`'s `index === pos` always matches and the `pos + 1` fallback is dead code for all real editing. The word helpers (`#wordStartLeft`/`#wordEndRight`) are grapheme-safe too.

Drop-in guard (passes — locks the invariant so a future cursor change can't silently break it):
```ts
const seg = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
const onBoundary = (line: string, col: number) =>
  col === 0 || col === line.length || [...seg.segment(line)].some((g) => g.index === col);

it('keeps the cursor on a grapheme boundary through editing', () => {
  const rich = ['x','😀','é','y','👨‍👩‍👧','🇦🇺','👍🏽','z'];
  const ops = ['left','right','home','end','ctrl+left','ctrl+right','backspace','delete','ctrl+backspace','ctrl+delete'] as const;
  const s = new EditorState();
  for (const value of rich) { s.handleKey({ type: 'char', value }); expect(onBoundary(s.lines[s.cursorLine] ?? '', s.cursorCol)).toBe(true); }
  s.handleKey({ type: 'home' });
  for (let i = 0; i < 30; i++) for (const type of ops) { s.handleKey({ type }); expect(onBoundary(s.lines[s.cursorLine] ?? '', s.cursorCol)).toBe(true); }
});
```

**One reachable edge (your call — it's obscure).** The battery found a single off-boundary case: insert a base char immediately before a pre-existing *orphan combining mark*. Then a following `delete` hits the fallback and splits the cluster:
```ts
const s = new EditorState({ lines: ['́'], cursorLine: 0, cursorCol: 0 }); // orphan combining acute
s.handleKey({ type: 'char', value: 'e' }); // line is "é" (one cluster), but cursor sits at col 1 (mid-cluster)
s.handleKey({ type: 'delete' });           // → "e"  — removed just the accent, split the cluster
```
You need an orphan combining mark before the cursor, so it's a corner of a corner — not gating. If you want it closed, the one-liner is making `graphemeBoundaryAfter` snap the way `Before` already does — `index + segment.length > pos` → return that end — which handles any off-boundary `pos` and makes the fallback truly dead.

Core's verified safe; guard test above locks it; the one edge is documented with a fix if you want it. Holding the stamp on that edge call.

bananabot9000:
**Correction to the fix note above.** The `graphemeBoundaryAfter` one-liner stops the `pos + 1` fallback from firing, but it does **not** fix the cluster split. `delete` does `line.slice(0, cursorCol) + line.slice(end)` — with `cursorCol = 1` (mid-cluster), `slice(0, 1)` already keeps the base without its accent regardless of what `end` is. So the symmetric helper doesn't help here.

The real root is the **insert** leaving the cursor mid-cluster (typing `e` before an orphan `́` → `é` with the cursor at col 1). The fix belongs there — snap the cursor to a grapheme boundary after a merging insert — or have `delete`/`backspace` clamp `cursorCol` to a cluster boundary before slicing.

Test that pins the root cause (fails today):
```ts
it('keeps the cursor on a grapheme boundary when inserting before a combining mark', () => {
  const seg = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
  const onBoundary = (line: string, col: number) =>
    col === 0 || col === line.length || [...seg.segment(line)].some((g) => g.index === col);
  const s = new EditorState({ lines: ['́'], cursorLine: 0, cursorCol: 0 });
  s.handleKey({ type: 'char', value: 'e' });
  expect(onBoundary(s.lines[s.cursorLine] ?? '', s.cursorCol)).toBe(true); // fails: cursor at col 1
});
```
Still obscure (needs an orphan combining mark before the cursor), so not gating — flagging only so the record's accurate.

bananabot9000:
**Deep grapheme/cursor-alignment hunt** — extended battery on the real `EditorState`. Normal editing is clean: **2,781 invariant checks across every key op + 40 vertical moves, cursor always on a grapheme boundary.** The bugs cluster in two spots.

### 1. A whole family — insert/paste that fuses forward (one root)
The combining-mark case from before is just the tip. `char` insert does `cursorCol += value.length`, which is wrong whenever the inserted tail merges with the *following* grapheme: the cursor lands mid-cluster, and a later `delete` then splits the cluster into broken surrogates / dropped joiners. Same root, five flavours — all ran on the real class, all reproduce:

| insert | before | becomes | `delete` then yields |
|---|---|---|---|
| `e` | `´` combining | `é` | `e` |
| `🇦` | `🇺` regional indicator | `🇦🇺` (flag) | `🇦` + lone half-regional |
| `👍` | `🏽` skin-tone | `👍🏽` | `👍` + lone modifier surrogate |
| `👨` | `‍👧` ZWJ+emoji | `👨‍👧` (family) | `👨👧` (ZWJ dropped) |
| `❤` | `️` VS16 | `❤️` | `❤` (VS16 dropped) |

It also hits **paste**: pasting `abce` before `´` lands the cursor mid-`é`. Paste is the realistic vector (pasted text ending in a char that fuses with existing content); the single-key cases are obscure individually but they are the same bug.

**One fix closes all of B + paste:** after a `char` insert, snap `cursorCol` to the grapheme boundary at/after the naive `cursorCol += value.length` (re-segment the line, clamp forward), so the cursor always lands *after* the full merged cluster. (Note: fixing `graphemeBoundaryAfter` does NOT help here — `delete`'s `slice(0, cursorCol)` splits at the mid-cluster col regardless. The root is the insert.)

### 2. No goal column on vertical movement (everyday UX)
Cursor at col 16, `up` through a 2-char line → lands col 2, and the original column is gone — the next `up` stays at col 2 instead of restoring 16. Standard editors keep a "preferred column" so up/down through short lines returns to the original. Separate fix: store a goal column, reset it on horizontal moves / edits.

### Guard test for the family (fails today)
```ts
it('keeps the cursor on a grapheme boundary after a fusing insert', () => {
  const onB = (l: string, c: number) =>
    c === 0 || c === l.length || [...new Intl.Segmenter(undefined, { granularity: 'grapheme' }).segment(l)].some((g) => g.index === c);
  for (const [line, ins] of [['́', 'e'], ['🇺', '🇦'], ['️', '❤']] as const) {
    const s = new EditorState({ lines: [line], cursorLine: 0, cursorCol: 0 });
    s.handleKey({ type: 'char', value: ins });
    expect(onB(s.lines[s.cursorLine] ?? '', s.cursorCol)).toBe(true);
  }
});
```

So: one root fix (insert-snap) closes the whole merge family + paste; a separate goal-column fix for the vertical drift. Batch those two and the next round is an approve.

bananabot9000:
Here's the family as an `it.each` matrix. Ran it against current HEAD (`dc2cfc4`) — **all 7 fail**, which matches what you're seeing: the delete-side fix (`11f871a`) doesn't touch the insert stranding the cursor mid-cluster, so the accent case (and the rest) persist.

```ts
import { describe, expect, it } from 'vitest';
import { EditorState } from '../src/model/EditorState.js';

const seg = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
const onBoundary = (line: string, col: number): boolean =>
  col === 0 || col === line.length || [...seg.segment(line)].some((g) => g.index === col);

describe('EditorState — a fusing insert must not strand the cursor mid-cluster', () => {
  it.each([
    { name: 'combining acute  (e + ´ -> é)',     before: '́',          insert: 'e' },
    { name: 'combining grave  (e + ` -> è)',     before: '̀',          insert: 'e' },
    { name: 'regional flag    (🇦 + 🇺)',         before: '\u{1F1FA}',       insert: '\u{1F1E6}' },
    { name: 'skin-tone        (👍 + 🏽)',         before: '\u{1F3FD}',       insert: '\u{1F44D}' },
    { name: 'ZWJ family       (👨 + ‍👧)',        before: '‍\u{1F467}', insert: '\u{1F468}' },
    { name: 'variation sel.   (❤ + VS16)',       before: '️',          insert: '❤' },
    { name: 'paste fusing tail (…e before ´)',   before: '́',          insert: 'abce' },
  ])('$name keeps the cursor on a grapheme boundary', ({ before, insert }) => {
    const s = new EditorState({ lines: [before], cursorLine: 0, cursorCol: 0 });
    s.handleKey({ type: 'char', value: insert });
    expect(onBoundary(s.lines[s.cursorLine] ?? '', s.cursorCol)).toBe(true);
  });
});
```

Current results on this branch:
```
combining acute (é)        cursorCol 1   FAIL   delete -> "e"
combining grave (è)        cursorCol 1   FAIL   delete -> "e"
regional flag   (🇦🇺)       cursorCol 2   FAIL   delete -> "🇦" + half-regional
skin-tone       (👍🏽)       cursorCol 2   FAIL   delete -> "👍" + lone surrogate
ZWJ family      (👨‍👧)       cursorCol 2   FAIL   delete -> "👨👧" (ZWJ dropped)
variation sel.  (❤️)        cursorCol 1   FAIL   delete -> "❤"
paste fusing tail           cursorCol 4   FAIL   ("abcé", cursor mid-é)
```

All 7 go green with the one insert-snap (clamp `cursorCol` forward to a grapheme boundary after a `char` insert). The delete-side fix can't reach them because `delete` slices `slice(0, cursorCol)` at the stranded mid-cluster col regardless.

bananabot9000 (approved):
Sanity check done — re-reviewed at current HEAD.

**The insert-snap fix (`efbc65c`) is correct and closes the family.** `char` now snaps `cursorCol` to `graphemeBoundaryAtOrAfter(next, cursorCol + value.length)`, landing after the full merged cluster. Re-ran the matrix against this HEAD: **7/7 pass** (combining acute/grave, regional flag, skin-tone, ZWJ, variation selector, paste) — the whole merge-insertion family is closed, cursor on a grapheme boundary every time.

**Ghost-text core intact after the main merge** — the three `#previous = null` reset sites (enter / resize / scroll), `disableAutowrap`, and `diffToWrites`' full-row absolute rewrite are all present. The cell-grid renderer (the mission) is unchanged and sound. Delete fix (`11f871a`) still grapheme-aware.

The remaining grapheme-first redesign + the vertical goal-column are scoped to #374, by agreement — out of this PR's mission.

Mission done and verified, nothing left to flag. Approving. ✅

PR #375, opened 2026-06-26
Own the API transport so retries are capped and abortable

shellicar:
## Summary

- Replace the @anthropic-ai/sdk runtime with an owned fetch/SSE transport (the SDK is kept for its types only)
- Cap and ESC-abort retry-after waits instead of honouring the uncapped server wait that froze the CLI
- Give up on a sustained account limit with a single notice rather than hanging

Needs manual API testing before merge: this rewrites how requests are issued and streamed, which build and type-check do not cover. Do not enable auto-merge.


bananabot9000:
Reviewed the retry/abort core — `backoff.ts`, `AccountLimitNotice`, and the `TurnRunner` loop. (Didn't exercise the SSE transport / token refresh — that's your manual API run; see the scope note.)

**The Achilles-heel fix is correct.** Walked the loop: an account-limit 429 (retry-after > the 60s cap, via `isAccountLimit`) never honours the server's wait — it sleeps the capped `RETRY_AFTER_CAP_MS` (60s) instead, so "Retry-After: 15 days" becomes a 60s nap, not a fortnight. The give-up is clean: `firstAccountLimitAt` anchors a 10-min wall-clock budget, checked *before* each wait, and on expiry it splices the single `stopped()` notice and throws `AccountLimitStoppedError`. Sustained limit → ~10 capped retries over 10 min → one loud notice → stop. No silent zombie.

**Abort handling is thorough — more so than exec-core was.** `defaultSleep` has the upfront `if (signal.aborted) return` guard *and* the abort listener (cancel timer + resolve), so an already-aborted or mid-sleep ESC cancels at once — exactly the guard #371 was missing. The loop also checks `abortSignal.aborted` at the top of the catch (in-flight cancel → rethrow) and after each sleep (`throwIfAborted`). ESC can't get stuck in a backoff.

Three minor notes, none gating:
1. For *transient* (sub-cap) 429s, the retry-after header is used only to *classify* (>cap → account-limit); the actual wait is exponential backoff, not the server's value. Standard and fine — just noting retry-after is classification-only on that path.
2. An account-limit 429 that arrives *without* a retry-after header falls to the transient path (10 backoff retries, then a raw 429 thrown) — it misses the `stopped()` notice. Fine if account-limit 429s always carry retry-after (they should); flagging the assumption.
3. The 10-min budget is per-*turn* (`firstAccountLimitAt` is loop-local), so each turn independently retries ~10 min before giving up. Reads intentional — just confirming it's not a session-wide budget.

Scope: I verified the retry / cap / abort / give-up *logic* (the stated purpose), and it's sound. I did not read `MessageStreamer` / `StreamProcessor` / `AnthropicClient` / token refresh — the SSE wire is what your manual API run covers, not a static read. First pass, no approval — and the real gate here is your wire test regardless. Happy to go deeper on the transport if you want a second set of eyes there.

bananabot9000:
Checked the SSE + tests now.

**SSE parser (`sse.ts`) is correct.** `parseFrame` handles the spec — field/value split on the first colon, single-leading-space strip, comment (`:`) lines skipped, multiple `data:` lines joined with `\n`, trailing `\r` stripped. `findFrameBoundary` picks the earliest of `\n\n` / `\r\n\r\n` with the right delimiter length. `parseSse` buffers across reads with a streaming `TextDecoder` (a multibyte char split across chunks is safe), slices complete frames, skips `ping`, throws `ApiStreamError` on `error`, and flushes a trailing frame that lacks a terminating blank line.

**Streaming-phase abort works** — the part that matters for "nothing hangs you." The transport does `AbortSignal.any([params.signal, timeout])` and passes it to `fetch`, so an ESC or the timeout aborts the fetch → the stream read rejects → `parseSse` throws → re-thrown as the abort reason. A slow/hung stream is escapable, not just the retry waits, and the timeout is the backstop if the user never hits ESC.

**Tests are comprehensive and hit the load-bearing paths directly** — the strong part:
- `backoff.spec`: `isAccountLimit` across all four edges (>cap, <cap, no-retry-after, non-429), and `defaultSleep` for *both* abort cases (already-aborted, fires-during-sleep).
- `TurnRunner.spec`: caps the wait at the cap *not* the header value (the Achilles-heel fix, tested), `AccountLimitStoppedError` on budget elapse, "does not wait after the give-up decision," `retrying`/`stopped` counts, and ESC-during-the-wait (throws the reason, does *not* raise `stopped`).
- `transport.spec`: HttpError/status/retry-after parsing, ConnectionError on non-abort reject, abort-reason on caller abort.
- `sse.spec`: frame split across two reads, CRLF-vs-LF priority, error-frame → ApiStreamError, trailing-flush, ping, comments.

That's exactly the coverage I'd want on the risky logic — the cap, the budget, the give-up ordering, both abort paths, and the wire framing.

Three minor, non-gating:
1. `parseSse(body, _signal)` takes the signal but doesn't use it — abort is handled at the fetch level (correctly), so the param is dead. Use it for an in-loop check or drop it; cosmetic.
2. The `finally` does `reader.releaseLock()` but not `reader.cancel()`. On a normal early-return (consumer stops without aborting), the underlying stream stays open. Rare path, but `cancel()` is the more robust close.
3. Data frames use raw `JSON.parse` (a bare, non-retryable `SyntaxError`) where the `error` frame uses `safeJsonParse`. A malformed data frame surfaces as a confusing hard error rather than a wrapped stream error. Asymmetry; minor.

Verdict: code and tests are in great shape — I'd sign off the *code* review on this. Holding the stamp only because the wire is the real gate and that's yours, not mine: live requests / SSE / token-refresh against the actual API aren't something a static read or these unit tests can prove. Once your manual run passes, say the word and I'll stamp the code review.

bananabot9000 (approved):
Verified the three fixes (`eb51eba`) — all clean:

1. **`_signal` dropped** from `parseSse`, and the `transport.ts` call updated to match. Abort stays correctly at the fetch level.
2. **`reader.cancel()` over `releaseLock()`** in the `finally`, with the rationale in a comment — and good catch on the test trap: a closed `ReadableStream`'s `cancel` is a spec no-op, so the test has to leave the stream *open* to actually prove the fix. That's exactly the kind of thing that passes a green-but-meaningless test; nice that you cornered it.
3. **Data-frame `JSON.parse` wrapped** → a corrupt frame now throws a classified `ApiStreamError` instead of a bare `SyntaxError`, so a broken stream is handled like any other stream failure.

That clears my notes, and the wire run (real requests / SSE / token refresh — Claude happily streaming) cleared the one gate a static read couldn't. Code and tests were sound across both passes; nothing left to flag.

Approving. ✅ The drone's airworthy.

PR #376, opened 2026-06-27
Add a persistent, relevance-searchable Memory tool for Claude

shellicar:
## Summary

- Claude writes a memory in one session and any later session, on any project, finds it by describing what it needs in plain words and gets it back ranked best first
- Relevance search runs on FTS5 with BM25, weighting title over keywords over body, so plain-words queries surface the most useful memories rather than substring matches
- Each memory is stamped with its origin (org, project and repo from the git remote, fleet from config) and its type, captured at write time and returned on every result
- Memories persist across restarts in a machine-wide SQLite store under ~/.claude
- Replaces the file-based testament, which forced one dated file to be many kinds of record at once


bananabot9000:
Read the engine (`SqliteMemoryEngine`), the FTS5 escaping (`toFtsMatch`), the git provider, and the engine tests. Didn't read `parseGitRemote` or the tool wiring — flagged below.

**The core is correct, and the riskiest part is the safe part.** A plain-words query hitting FTS5 `MATCH` is exactly where I'd expect a crash or an injection, and it's handled: `toFtsMatch` keeps only `\p{L}\p{N}` runs, double-quotes each token, OR-joins, returns null on empty. Every operator (`-`, `*`, quotes, parens, even a literal `OR`) is stripped or neutralised into a quoted term — and it's *tested* (the input-safety cases, including `hello -world` not being a NOT). BM25 weights are right too: `10.0, 1.0, 4.0` maps to columns `title, body, keywords` → title > keywords > body, matching the stated intent and the ranking test. WAL + `busy_timeout` + soft-delete are sound for a machine-wide shared store.

Real, minor, honest severity:
1. **id lookups are full scans.** `read`/`delete` do `WHERE id = ?` on `id`, which is `UNINDEXED` in the FTS5 table — so each is a linear scan. Fine at hundreds of rows; degrades as the store grows. A side id→rowid map fixes it *if* id-lookup ever gets hot. Not a now-problem.
2. **Soft-deleted rows stay in the FTS index.** Correctly filtered from results (`deleted_at IS NULL`), but they linger and nudge the BM25 corpus stats (IDF). Minor; a periodic rebuild handles it if it ever matters.
3. **Search is broad-recall by design.** OR-joining tokens means a 5-word query matches anything with *one* of the five, ranked by BM25 — the intended plain-words behaviour, not a bug. Confirming it's deliberate: no query requires all terms.
4. **Didn't read `parseGitRemote`.** The git-URL → org/project/repo parse is the one spot I'd scrutinise next (ssh vs https, nested groups, trailing `.git` are classic edge cases). No finding — flagging the gap honestly.

First pass — no approval (didn't read the wiring/permissions; first pass is review-only). But the substance is clean. And for the record: I'm not relitigating the in-memory-fake-vs-SQLite thing from the plan. `RecordingMemoryStore` is a test stub; it owes the engine nothing. The SQLite is correct — that's the part that matters.

shellicar:
> 1. **id lookups are full scans.** `read`/`delete` do `WHERE id = ?` on `id`, which is `UNINDEXED` in the FTS5 table — so each is a linear scan. Fine at hundreds of rows; degrades as the store grows. A side id→rowid map fixes it _if_ id-lookup ever gets hot. Not a now-problem.
> 2. **Soft-deleted rows stay in the FTS index.** Correctly filtered from results (`deleted_at IS NULL`), but they linger and nudge the BM25 corpus stats (IDF). Minor; a periodic rebuild handles it if it ever matters.

Thanks, making a change for both of these

> 3. **Search is broad-recall by design.** OR-joining tokens means a 5-word query matches anything with _one_ of the five, ranked by BM25 — the intended plain-words behaviour, not a bug. Confirming it's deliberate: no query requires all terms.

Intentional.

> 4. **Didn't read `parseGitRemote`.** The git-URL → org/project/repo parse is the one spot I'd scrutinise next (ssh vs https, nested groups, trailing `.git` are classic edge cases). No finding — flagging the gap honestly.

Works fine, in the future you should probably read it though rather than speculating.

At least check the tests, because regardless of how the code works, the tests show what it's designed to support.

bananabot9000 (commented):
**Review — current head (`88d52eca`).** Read the files, not just the diff. Both prior findings + the rowid nit are resolved; one latent concurrency issue to flag.

✅ **Verified fixed:**
- **id full-scan → indexed.** `memory_index (id PK, fts_rowid)` + `read`/`delete`/`archive` JOIN through it. No scan.
- **The rowid nit** — the column is now `fts_rowid`, not the shadow-prone `rowid`. 👍
- **Soft-delete linger → archive.** No `deleted_at` in the FTS `memories` table; `delete()` does archive-copy → delete-row → delete-index atomically under `BEGIN IMMEDIATE`, idempotent on an unknown id. The bm25 corpus never sees tombstones.
- `node:sqlite` (no native-binding / ABI build risk); `CREATE VIRTUAL TABLE … fts5` doubles as the eager FTS5-availability check. `migrate.ts` lockout is correct — newer-major refused, newer-minor tolerated-not-migrated (the mixed-version coexistence stays intact).

🔶 **One latent finding — migration concurrency (your call; not ship-blocking today):**
`migrate()` reads `user_version` **once, before** the per-migration `BEGIN IMMEDIATE`, and never re-checks inside the lock. Two CLIs starting on the same store (the scenario `busy_timeout` exists for) both compute the same `pending` list; the lock serialises them, but the race-loser still **re-applies** every migration it pre-computed.

Harmless *today* only because migration 1 is fully idempotent (`CREATE … IF NOT EXISTS`, `INSERT OR IGNORE`). The first **non-idempotent** migration — e.g. `ALTER TABLE … ADD COLUMN` — double-applies under that race: the second CLI throws `duplicate column name`, rolls back, fails to start. A real failure mode in exactly the multi-CLI design this targets, deferred to the next schema change.

Fix is a few lines: inside each `BEGIN IMMEDIATE`, re-read `user_version` and `continue` if it's already ≥ this migration's version, so the loser no-ops instead of re-applying.

No approval this pass — one genuine thing to say. Not blocking the memory tool: it can't bite until a non-idempotent migration lands. Guard it now, or file as a follow-up gated on the next schema bump, and I'll stamp.

bananabot9000 (approved):
**Verified & approved** (`a2b2620`). Re-read `migrate()` and traced the race: each migration's `BEGIN IMMEDIATE` now re-checks `currentVersion(db)` under the lock and only applies if still ahead. The 50-way tmux-restore resolves cleanly — the winner applies once and bumps the version; all 49 losers take the lock, see it already bumped, skip, and commit empty. A non-idempotent `ALTER TABLE` applies exactly once; nothing double-applies, nothing throws `duplicate column`. It also handles partial multi-step races correctly (per-migration lock + re-check, not a single outer check).

That was the last open item — id index, `fts_rowid` rename, atomic archive-delete, and now race-safe migrations all resolved. CI green.

Optional, not gating: there's no concurrency *test* — the fix is correct by the lock semantics, but a two-connection race test would prove it empirically. Happy to add one if you want it nailed down rather than reasoned.

PR #377, opened 2026-06-27
Repaint every TUI row each frame to survive external grid mutation

shellicar:
## Summary

- Repaint every TUI cell each frame instead of diffing against the previous grid
- Cause-agnostic safety net: an externally mutated cell (e.g. tmux reflow on resize or copy-mode) can no longer outlive a single paint
- Robustness insurance, not a bug fix: the orphaned-character ghosting it guards against was an external tmux issue (3.6b, fixed upstream by 3.7), not a defect in our renderer

### DIFF
--- a/.claude/testament/2026-06-27.md
+++ b/.claude/testament/2026-06-27.md
@@ -1,0 +1,48 @@
+# 19:24
+
+## repaint-every-cell — Phase 1 Maker (iteration 1)
+
+Killed the orphaned-character ghosting by replacing the row-diff in
+`ScreenBuffer.diffToWrites` with a full repaint: every row, every frame, from
+column 1. The previous-grid skip was the bug — under tmux the grid is reflowed
+behind the renderer's back (resize, copy-mode entry), so a row our model
+believed unchanged could still hold orphaned cells, and the skip kept them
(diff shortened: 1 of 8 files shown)


bananabot9000 (approved):
Approved. Read the full diff including the proving test.

The change is sound and — the part that matters — honestly framed: robustness insurance (every cell repainted so an externally-mutated cell can't outlive a frame), not a bug fix; `changed`, not `fixed`, because the ghost was an upstream tmux 3.6b issue (fixed by 3.7), not a renderer defect.

Verified:
- `diffToWrites` drops the row-skip and repaints every row from column 1. Sound; `_prev` is now a dead param kept for call-site compat (non-gating — clean up whenever).
- `repaintEveryCell.spec.ts` proves it for the right reason: paint, `poke` a `Z` onto an otherwise-quiet row, repaint the *same* frame — the old skip would keep the Z, the full repaint scrubs it. Real renderer + `GridScreen`, not a logic mock.
- Correctly landed at state 2 (row repaint), not the per-cell variant that dropped bytes.

The 2026-06-28 testament is the standout — a complete 'it was tmux, here's the proof, don't re-chase it' record with a minimal upstream repro. Worth more than the code.

PR #378, opened 2026-06-28
Adopt core-di-lite property injection across every package

shellicar:
## Summary

- Convert every resolved dependency to core-di-lite property injection and delete the hand-wired constructor factories
- Split config loading into a pure read, a holder, and a watch handle with no load or start step
- Create SQLite databases through a registered factory with an in-memory path for verification
- Move CLI startup into main() so the entry module's only import-time effect is invoking it
- Pin vitest to Vite 7, since oxc on Vite 8 cannot lower the standard decorators


bananabot9000 (commented):
**Deeper pass — self-configuration review** (criterion: a class shouldn't configure itself; setup that can fatally fail should surface eagerly at `buildProvider`, so a resolve tells you whether SQLite/migrations work).

**The real one — `DurableConfigFactory`.** The constructor ends with `this.config = this.#build()`, and `#build()` reads `this.#configLoader.config` — the class reaches into config to configure itself, in its constructor, rather than being handed a finished value. Every other config consumer avoids this: the `IObjectStore` / `PermissionsNoticeGate` / `StatusState` factories *extract* the config value and pass the finished thing. Worth deciding whether `#build` belongs in the container factory (resolve config → build → pass a finished `DurableConfig`), leaving the class a holder.
- Minor, same file: `resolveSystemPromptsFor`'s doc comment calls itself "activation" — the exact word the testament flags as the relabel smell. The call is fine (genuine async file I/O over a constructed object, same category as `ts.start`); only the wording brushes the rule.

**Checked and clean:**
- `SqliteObjectStore` — does **not** use the database name or create the db object. It is handed a `DatabaseSync` and owns only its own schema (`CREATE TABLE` + prepared statements) on the given resource; name resolution and db creation are the `IObjectStore` factory's job. The `CREATE TABLE` runs eagerly when that factory resolves at `buildProvider`, so a migration that cannot apply fails at boot — the eager-resolution guarantee, not a violation.
- `ToolRegistry` — builds its wire-map from the *given* tools (self-init of its own index, not from config); eager-by-design so a bad schema surfaces at boot.
- `ConfigLoader` (the holder model), `AppToolsService`, `DatabaseFactory` (db creation lives in `getDatabase`/`#open`, correctly the factory's job), `AnthropicClient`.

Net: composition is correctly in the container factories; the one place a class reaches into config to build itself is `DurableConfigFactory.#build`.

apps/claude-sdk-cli/test/AccountLimitNotice.spec.ts
```diff
@@ -15,14 +18,25 @@ class RecordingConversationState extends ConversationState {
   }
 }
 
+// AccountLimitNotice injects ConversationState, so build it through a container
+// holding the provided (recording) state.
+function buildAccountLimitNotice(conversation: ConversationState): AccountLimitNotice {
+  const services = createServiceCollection();
+  // ConversationState @dependsOn(IClockProvider); buildProvider injects it eagerly.
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Should be using a fake or mock clock.

apps/claude-sdk-cli/test/AgentMessageHandler.spec.ts
```diff
@@ -43,40 +50,74 @@ type OptsOverrides = {
...
+  const durableConfig = makeConfig(overrides.config);
+  const configLoader = {
+    get config() {
+      return fullConfig;
+    },
+  } as unknown as ConfigLoader<any>;
+  const appTools = { tools: durableConfig.tools, store, refTransform: (_name: string, output: unknown) => output } satisfies AppToolsService;
+
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Same here

apps/claude-sdk-cli/test/CommandIntentExecutor.spec.ts
```diff
@@ -22,7 +26,20 @@ function makeExecutor(source: AttachmentSource) {
       cycleCalls.effort += 1;
     },
   };
-  const executor = new CommandIntentExecutor(commandModeState, conversationState, session, source, modelSettings);
+  const services = createServiceCollection();
+  services.register(CommandModeState).to(CommandModeState, () => commandModeState);
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: And here

apps/claude-sdk-cli/test/CommandKeyHandler.spec.ts
```diff
@@ -25,8 +30,18 @@ function makeHandler(sourceText: string | null = null) {
       cycleCalls.effort += 1;
     },
   };
-  const executor = new CommandIntentExecutor(commandModeState, conversationState, session, source, modelSettings);
-  const handler = new CommandKeyHandler(commandModeState, COMMAND_BINDINGS_BY_CONTEXT, executor);
+  const services = createServiceCollection();
+  services.register(CommandModeState).to(CommandModeState, () => commandModeState);
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Another one

apps/claude-sdk-cli/test/ConversationState.spec.ts
```diff
@@ -1,7 +1,18 @@
...
+import { IClockProvider } from '@shellicar/claude-core/providers/IClockProvider';
+import { createServiceCollection } from '@shellicar/core-di-lite';
 import { describe, expect, it } from 'vitest';
 import { ConversationState } from '../src/model/ConversationState.js';
 
+// ConversationState injects IClockProvider; build it through a container wrapping
+// the supplied clock (defaulting to the system clock).
+function buildConversationState(clock: Clock = Clock.systemUTC()): ConversationState {
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock }));
```
shellicar: Bingo

apps/claude-sdk-cli/test/HistoryNavHandler.spec.ts
```diff
@@ -1,9 +1,23 @@
...
 import { describe, expect, it } from 'vitest';
 import { HistoryNavHandler } from '../src/controller/HistoryNavHandler.js';
 import { ConversationState } from '../src/model/ConversationState.js';
 import { HistoryViewState } from '../src/model/HistoryViewState.js';
 import { TerminalState } from '../src/model/TerminalState.js';
 
+// HistoryNavHandler injects HistoryViewState/ConversationState/TerminalState; build it through a container.
+function buildHistoryNavHandler(state: HistoryViewState, conversation: ConversationState, terminal: TerminalState): HistoryNavHandler {
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Yahtzee

apps/claude-sdk-cli/test/renderConversation.spec.ts
```diff
@@ -1,8 +1,19 @@
...
+import { createServiceCollection } from '@shellicar/core-di-lite';
 import stringWidth from 'string-width';
 import { describe, expect, it } from 'vitest';
 import { ConversationState } from '../src/model/ConversationState.js';
 import { buildDivider, type DividerTimestamps, renderConversation } from '../src/view/renderConversation.js';
 
+// ConversationState injects IClockProvider; build it through a container.
+function buildConversationState(): ConversationState {
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Here

apps/claude-sdk-cli/test/StoreEmissions.spec.ts
```diff
@@ -1,11 +1,21 @@
...
 import { ConversationState } from '../src/model/ConversationState.js';
 import { EditorState } from '../src/model/EditorState.js';
 import { StatusState } from '../src/model/StatusState.js';
 import { ToolApprovalState } from '../src/model/ToolApprovalState.js';
-import { MemoryFileSystem } from './MemoryFileSystem.js';
+
+// ConversationState injects IClockProvider; build it through a container.
+function buildConversationState(): ConversationState {
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: .

apps/claude-sdk-cli/test/ThinkingRequestWiring.spec.ts
```diff
@@ -45,34 +54,77 @@ class FakeMessageStreamer extends IMessageStreamer {
...
+// TurnRunner is property-injected; build it through a container with test doubles.
+function buildTurnRunner(streamer: IMessageStreamer): TurnRunner {
+  const services = createServiceCollection();
+  services.register(IMessageStreamer).to(IMessageStreamer, () => streamer);
+  services.register(IStreamProcessor).to(StreamProcessor);
+  services.register(ILogger).to(NoopLogger);
+  services.register(AccountLimitListener).to(NoopAccountLimitListener);
+  services.register(ISleepProvider).to(ISleepProvider, () => ({ sleep: async () => {} }));
+  services.register(IRandomProvider).to(IRandomProvider, () => ({ next: () => 0.5 }));
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar:
```diff
- services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
+ services.register(Clock).to(Clock, () => Clock.systemUTC());
```

apps/claude-sdk-cli/test/ViewHost.spec.ts
```diff
@@ -165,13 +183,25 @@ describe('ViewHost — escape routing through the primary chains', () => {
...
     const cancelLog: string[] = [];
-    const executor = new CommandIntentExecutor(model.commandModeState, model.conversationState, model.session, new FakeAttachmentSource(), { cycleThinking: () => {}, cycleEffort: () => {} });
-    const approvalHandler = new ApprovalHandler(model.toolApprovalState);
-    const commandKeyHandler = new CommandKeyHandler(model.commandModeState, COMMAND_BINDINGS_BY_CONTEXT, executor);
-    const editorHandler = new EditorHandler(model.editorState, model.commandModeState, model.terminalState);
-    const cancelHandler = new CancelHandler(() => cancelLog.push('cancel'));
-    const editorChain: readonly InputHandler[] = [approvalHandler, commandKeyHandler, editorHandler];
-    const streamingChain: readonly InputHandler[] = [approvalHandler, cancelHandler];
+    const services = createServiceCollection();
+    services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: I lost count

apps/claude-sdk-cli/test/ViewSelectHandler.spec.ts
```diff
@@ -1,9 +1,23 @@
...
 import { describe, expect, it } from 'vitest';
 import { ViewSelectHandler } from '../src/controller/ViewSelectHandler.js';
 import { AppModeState } from '../src/model/AppModeState.js';
 import { ConversationState } from '../src/model/ConversationState.js';
 import { HistoryViewState } from '../src/model/HistoryViewState.js';
 
+// ViewSelectHandler injects AppModeState/HistoryViewState/ConversationState; build it through a container.
+function buildViewSelectHandler(appModeState: AppModeState, historyViewState: HistoryViewState, conversation: ConversationState): ViewSelectHandler {
+  const services = createServiceCollection();
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: Clock.systemUTC() }));
```
shellicar: Ding

packages/claude-core/src/providers/SystemClockProvider.ts
```diff
@@ -0,0 +1,8 @@
+import { Clock } from '@js-joda/core';
+import { IClockProvider } from './IClockProvider';
+
+export class SystemClockProvider extends IClockProvider {
```
shellicar: See my comment in `container.ts`

packages/claude-sdk/test/TurnRunner.spec.ts
```diff
@@ -134,6 +140,33 @@ function accountLimitError(): HttpError {
...
+ */
+function buildTurnRunner(streamer: IMessageStreamer, processor: IStreamProcessor, logger?: ILogger, listener?: AccountLimitListener, sleep?: (ms: number, signal: AbortSignal) => Promise<void>, random?: () => number, clock?: Clock): TurnRunner {
+  const services = createServiceCollection();
+  services.register(IMessageStreamer).to(IMessageStreamer, () => streamer);
+  services.register(IStreamProcessor).to(IStreamProcessor, () => processor);
+  services.register(ILogger).to(ILogger, () => logger ?? new NoopLogger());
+  services.register(AccountLimitListener).to(AccountLimitListener, () => listener ?? new SpyListener());
+  services.register(ISleepProvider).to(ISleepProvider, () => ({ sleep: sleep ?? (async () => {}) }));
+  services.register(IRandomProvider).to(IRandomProvider, () => ({ next: random ?? (() => Math.random()) }));
+  services.register(IClockProvider).to(IClockProvider, () => ({ clock: clock ?? Clock.systemUTC() }));
```
shellicar: Beep

packages/claude-core/src/providers/TimeoutSleepProvider.ts
```diff
@@ -0,0 +1,20 @@
...
+    }
+    return new Promise<void>((resolve) => {
+      const handle = setTimeout(resolve, ms);
+      signal.addEventListener(
+        'abort',
+        () => {
+          clearTimeout(handle);
+          resolve();
+        },
+        { once: true },
```
shellicar: { once: true } only auto-removes it after it fires, i.e. after an abort.
If the same signal is reused they pile up.
MaxListenersExceededWarning at 10, plus the retained closures.

shellicar:
> **The real one — `DurableConfigFactory`.** The constructor ends with `this.config = this.#build()`, and `#build()` reads `this.#configLoader.config` — the class reaches into config to configure itself, in its constructor, rather than being handed a finished value. Every other config consumer avoids this: the `IObjectStore` / `PermissionsNoticeGate` / `StatusState` factories _extract_ the config value and pass the finished thing. Worth deciding whether `#build` belongs in the container factory (resolve config → build → pass a finished `DurableConfig`), leaving the class a holder.

This one was hard to spot since there's no diff, but inspecting it closer revealed a lot.
It is building the config, but it was duck-typed to an interface and circumvented it's own readonly field using Object.assign for no reason.

bananabot9000 (commented):
**Second pass — changes since my deeper-pass review.** Read the four commits since `981bab13`. One thing to flag, two confirmed resolved.

🔶 **The Clock swap flips the per-turn time stamp to UTC.** `container.ts` now registers `Clock` as `Clock.systemUTC()`, where the deleted `SystemClockProvider` returned `Clock.systemDefaultZone()`. Most consumers are unaffected — `ConversationState` and `TurnRunner` take `Instant.now(clock)` / `clock.instant()`, which are zone-independent. But `TurnRunner.ts:71` passes the injected clock into `formatClockStamp(this.clock)`, and that formats a **zoned** `ZonedDateTime.now(clock)`. So the human-readable "what time is it" reminder injected each turn now renders in UTC instead of local — the docstring example is literally `Australia/Melbourne (+10:00)`.

It's easy to miss because `clockStamp.ts`'s default param still says `Clock.systemDefaultZone()` — the function *looks* local; only the injected call site overrides it. The testament notes the wrapper removal but not the zone change, so I'm flagging it as possibly inadvertent. If the stamp should stay local, either register `systemDefaultZone()` or have `formatClockStamp` apply the local/configured zone instead of trusting the injected clock's zone. If UTC-in-the-stamp is intended, ignore me.

✅ **`DurableConfigFactory` self-config flag — resolved well.** Constructor work is gone (property injection), and `config` is a derive-on-read getter returning a fresh object — which also removes the old `Object.assign(this.config, …)` in-place mutation, so consumers holding the reference no longer alias a mutated object. Container simplified to interface-only (`register(IDurableConfigProvider).to(DurableConfigFactory)`); I checked that nothing resolves the concrete `DurableConfigFactory` or `@dependsOn`s it anymore, so dropping the standalone registration + alias is safe.

✅ **`TimeoutSleepProvider` listener leak — fixed.** Both paths detach now (timeout branch `removeEventListener`, abort branch `{ once: true }`). Verified.

No approval on this pass — the clock-zone question is the open one, your call whether it's intended.

bananabot9000 (approved):
**Verified — clock-zone fix is correct and complete.** Injected `Clock` reverted to `systemDefaultZone()` (restores the pre-PR local behaviour, so the per-turn stamp reads local again), and `formatClockStamp` drops the default param so the clock is now a required argument — no more local-looking decoy hiding an injected override. Both callers (`TurnRunner`, the spec) already pass an arg, so nothing breaks; `type Clock` import is right now the value's unused; instants are zone-independent so unaffected either way.

That was the last open item. Across the passes: the `DurableConfigFactory` self-config flag (resolved — derive-on-read getter, interface-only registration verified safe), the `TimeoutSleepProvider` listener leak (both paths detach), and this clock zone — all resolved and verified. CI green. Approving.

PR #379, opened 2026-06-29
Replace the stale user-facing documentation

shellicar:
## Summary

- Rewrite the root README into the canonical full documentation: what the CLI is, why you would use it, and how to install, configure, run, and extend it
- Add READMEs for the CLI, the standalone mcp-exec server, and the libraries, so the empty npm pages now describe their packages
- Delete docs describing the removed legacy app and the stale cli-config schema
- Default the config to claude-opus-4-8 with high thinking effort
- Record the documentation change in each affected package changelog


bananabot9000 (approved):
**Verified & approved** (`b9302b23`). Spot-checked the docs against the code rather than stamping: every config key the root README documents — `model`, `maxTokens`, `thinking.enabled`/`effort`, `tools.execV2`, `serverTools.webSearch`/`webFetch`, `historyReplay.showThinking` — exists in `schema.ts`, and the documented default (`claude-opus-4-8`, effort `high`) matches the schema's real default plus the `DEFAULT_MODEL` constant. No invented or stale keys — the right bar for a docs PR whose job is killing staleness.

My one earlier concern — the `sonnet-4-6` → `opus-4-8` default flip riding in a docs PR — is resolved: it's docs and defaults brought into alignment in one pass (a README declaring the canonical config while the code defaulted to something else would be born stale), and it's disclosed in the body + changelog. Intentional, not a rider.

CI green. Approving.

PR #380, opened 2026-06-29
Add ExecV3 structured execution tool

shellicar:
## Summary

- Add ExecV3: a flat command list joined by forward-pointing operators, no shell
- Evaluate under bash-exact list semantics (`&&`/`||`/`|` precedence, exit status) with position-aligned per-command results
- Surface a call-time `intent` field and bash-equivalent example pairs the model reads at invocation
- Enable ExecV3 by default; ExecV2 now sits behind its flag
- Consolidate the SDK's tool-dispatch results into one outcome type with a new `ToolRefusedError`; a blocked program now refuses cleanly rather than returning a fabricated result


bananabot9000 (commented):
**Review — `afa75754`.** Strong tool; one red-build blocker.

✅ **Verified correct (read the semantics, not just the diff):**
- **bash list semantics** (`group.ts` + `engine.ts`). Traced `a && b || c`, `a || b && c`, and `a && b ; c`: equal-precedence left-associative `&&`/`||`, short-circuit, the sequential-runs-after-a-short-circuited-chain case, and exit-status threading all match bash exactly. Pipeline exit = last command (bash default, no pipefail). The hard part is right.
- **`runPipeline`**: stages run concurrently, bridged by `PassThrough`, with per-stage capture vs downstream-pipe vs file-redirect resolved correctly (and `| + stdout-redirect` rejected at validation). Per-command exit/signal preserved.
- 59 ExecV3 scenario tests + 5 group tests — good coverage of the semantics.

🔴 **Blocker — build (24.x) is red on one stale test:**
`test/Exec.spec.ts > Exec — cancellation > "rejects with ToolCancelledError when cancelled mid-execution"` throws a `ZodError` instead of the expected `ToolCancelledError`:
- `intent`: expected string, received undefined
- `description`: unrecognized key

That test still constructs the **old** input shape (`description`, no `intent`); the renamed schema (intent now required, description removed) rejects it during validation, so it never reaches the cancellation path it's meant to exercise. The production rename looks intentional — the fix is updating that one fixture (`description` → `intent`). (If `intent` was actually meant to be *optional*, it's the schema instead — but the PR reads like required-intent is the point.)

No approval — red build plus a real, if one-line, finding. Green it and I'll stamp; the engine and pipeline are already verified.

packages/claude-sdk-tools/src/ExecV3/runPipeline.ts
```diff
@@ -0,0 +1,85 @@
...
+      ([status, out, err]): CommandResult => ({
+        stdout: out,
+        stderr: err,
+        exitCode: status.exitCode,
+        signal: status.signal,
+      }),
+    );
+  });
+
+  return Promise.all(runs);
```
shellicar, line 82: A pipeline deadlocks when a downstream stage exits before draining its input. `yes | head -n 1` never returns, and the same goes for any `cmd | head` or `cmd | grep -q` where the consumer stops reading early.

The producer keeps writing into the bridge, the consumer is gone, and nothing tears the producer down. exec-core's `closeSinks` then waits on `finished()` for that undrained bridge, which never ends. The `timeout` does not rescue it: a 2s timeout was still hung at 6s.

This is the single-run model in exec-core, shared with V1 and V2. Fixing it properly means changing how a pipe sink is closed, either destroying the bridge or propagating the consumer's exit upstream. That is too big for this PR, so I'm recording it here rather than fixing it in this change.

packages/claude-sdk-tools/src/ExecV3/schema.ts
```diff
@@ -0,0 +1,168 @@
...
+            { program: 'pnpm', args: ['build'], op: '&&' },
+            { program: 'pnpm', args: ['test'] },
+          ],
+        ],
+      }),
+    timeout: z
+      .number()
+      .max(600000)
+      .default(30000)
+      .describe('Timeout in ms (default 30000, max 600000). Always enforced.')
```
shellicar: The schema describes `timeout` as "always enforced". On the deadlock above it never fires, so the claim is false where it matters most. Drop the line until the deadlock is fixed.

```diff
-    .describe('Timeout in ms (default 30000, max 600000). Always enforced.')
+    .describe('Timeout in ms (default 30000, max 600000).')
```

Same wording in `schema.md`.

shellicar (commented):
Three findings from reviewing ExecV3. The pipe deadlock is the real one. It is a change to exec-core's single-run model, so it is out of scope for this PR and recorded rather than fixed here. The other two are small: the `timeout` describe overstates what it enforces, and there is no test for the early-consumer-exit case that hides the deadlock.

packages/claude-sdk-tools/test/ExecV3/scenarios.spec.ts
```diff
@@ -0,0 +1,634 @@
...
+    const actual = result.success;
+    expect(actual).toBe(expected);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// pipe 2-stage — bash: echo hello | cat
+// ---------------------------------------------------------------------------
+
+describe('pipe — echo hello | cat', () => {
```
shellicar, line 295: No test covers a pipe whose consumer exits before draining, which is why the deadlock got through a green suite. The pipe scenarios all drain fully (`echo | cat`, `printf | grep | wc`). Add a `cmd | head` scenario that asserts a prompt return. It will hang until the deadlock is fixed, so it lands with that fix, marked skipped or expected-fail until then.

bananabot9000 (commented):
**Review — `5ae15d98`.** Judged the fixes against behaviour, not against the findings list.

✅ **C2 (timeout overclaim) — fixed honestly.** Dropped the false "always enforced" wording + `min(1)` lower bound. Correct: the timeout can't enforce on the deadlock path, so the right move is to stop promising it.

✅ **C3 (missing test) — and the right *way* to defer.** `it.fails` with the exact `yes | head -n 1` repro, 2 s-bounded so it can't hang CI. Green now, and it auto-flips to red the moment the pipe lifecycle is fixed — the deferred bug literally can't be forgotten.

✅ **C1 (pipe deadlock) — deferral is legitimate.** It's the exec-core rewrite (a single-command `run()` unit can't own an inter-command pipe lifecycle), parity with V1/V2/mcp-exec which share the mechanism, and the guard above pins it.

🔶 **The one open thing — the redirect-cwd finding fell through, untracked.** `resolveStageSinks` still does `createWriteStream(redirect.stdout, …)`, and `normalise` only expands `~`/`$VAR` — neither resolves a *relative* redirect path against the command's `cwd`. So `{ program: 'echo', cwd: '~/foo', redirect: { stdout: 'out.txt' } }` writes `out.txt` to the *process* cwd, not `~/foo`; bash writes it in the command's cwd. Silent wrong-location write, ExecV3-local (not exec-core), and — unlike the deadlock — nothing pins it. Deliberate defer, or did it slip? If deferred, it wants the same `it.fails` guard the deadlock got, or it's exactly the finding that goes crickets.

No approval — that one's the open decision. Once it's fix-or-defer-with-a-guard I'll stamp; the other three are sound.

packages/claude-sdk-tools/src/ExecV3/runPipeline.ts
```diff
@@ -0,0 +1,85 @@
...
+ * present and no stdout redirect diverts it, stdout flows onward and is NOT captured.
+ */
+function resolveStageSinks(cmd: Command, downstream: Writable | undefined): StageSinks {
+  const redirect = cmd.redirect;
+  const mergeStderr = redirect?.stderr === '&1';
+
+  let stdout: Writable;
+  let stdoutCapture: PassThrough | undefined;
+  if (redirect?.stdout != null) {
+    const file = createWriteStream(redirect.stdout, { flags: 'w' });
```
shellicar: A relative redirect path is written to the wrong directory when the command sets its own `cwd`. `normalise` only expands `~`/`$VAR`, so a relative path stays relative, and `createWriteStream` here resolves it against the node process cwd while the child runs in `cmd.cwd`. They agree only when the command has no `cwd`.

Proven: `echo hi` with `cwd` a temp dir and `redirect.stdout: 'out.log'` writes the file into the process cwd, not the temp dir. The stderr redirect just below (line 46) has the same fault.

Fixed in this change: both redirect paths now resolve against `cmd.cwd ?? ctx.cwd`, with a test pinning where each lands.

bananabot9000 (commented):
## Pipe early-exit doesn't short-circuit — hangs to the timeout

**Verified by running the mechanism**, because I had this exact area wrong before by theorising.

When a downstream stage exits before an upstream stage finishes — `yes | head`, `find / | head -n5`, `cat big | grep -m1 x` — the upstream keeps running. In `runPipeline`, stage *i*'s stdout is piped into a `PassThrough` bridge; when the reader stage exits, the bridge is never torn down, so the upstream process gets no SIGPIPE-equivalent and never exits. Its `executor.run(...)` never resolves on its own, and `Promise.all(runs)` waits on it.

**Repro (raw `spawn` + `PassThrough`, mirroring `runPipeline`'s shape):**
```
s1 (head) exit code=0    // downstream done, 5 lines captured
after 4s: s0 (yes) exited=false   // upstream still running
```

**Honest boundedness:** this is *not* an infinite deadlock — the `fromStream` rewrite dropped the old `finished(bridge)` that held the promise past the kill, and `execSignal(signal, input.timeout)` feeds the timeout into `ctx.signal`, so when `input.timeout` fires the upstream is aborted and the run resolves. So the failure mode is: **`yes | head` returns only after the full timeout, not when `head` is done.** For a real pipeline that means the whole timeout window is spent, and the upstream does all its wasted work (reads the whole tree, generates forever) before being cut off — the early-exit optimisation the pipe is *for* doesn't happen.

**Fix direction (yours to design):** propagate downstream completion upstream — when the last/consuming stage resolves, tear down the bridge and abort the earlier stages (the SIGPIPE analogue). Caveat: my repro is raw `spawn`, not exec-core's `Executor`; I've proven the bridge mechanism and traced the timeout path, but not that the `Executor` has no other teardown — worth a real `yes | head` test in the suite to pin it.

*(Not re-flagging the BLOCKED single-element `results` — that's pre-execution rejection, documented, intentional; I got that wrong last time.)*

bananabot9000:
## Runnable repro for the early-exit finding

**Important caveat on what this is:** I could not run the actual ExecV3 in my checkout (no `node_modules`). This is a **standalone mimic** of `runPipeline`'s shape — raw `child_process.spawn` + `PassThrough` bridge — so it proves the *bridge mechanism* (a `PassThrough` between two spawned processes does not propagate the downstream exit back to the upstream), not that ExecV3 as a whole has no other teardown. The real test belongs in the suite (below).

**`bridge-repro.mjs`** (no deps, `node bridge-repro.mjs`):
```js
import { spawn } from 'node:child_process';
import { PassThrough } from 'node:stream';

// mirrors runPipeline for `yes | head -n 5`:
// s0 stdout -> bridge -> s1 stdin ; s1 stdout -> capture
const bridge = new PassThrough();
const s0 = spawn('yes', { stdio: ['ignore', 'pipe', 'ignore'] });
s0.stdout.pipe(bridge);
const s1 = spawn('head', ['-n', '5'], { stdio: ['pipe', 'pipe', 'ignore'] });
bridge.pipe(s1.stdin);
let out = '';
s1.stdout.on('data', d => { out += d; });
s1.stdin.on('error', () => {}); // EPIPE if head closes stdin early
const exited = { s0: false, s1: false };
s0.on('exit', (c, sg) => { exited.s0 = true; console.log(`s0 (yes) exit code=${c} sig=${sg}`); });
s1.on('exit', (c, sg) => { exited.s1 = true; console.log(`s1 (head) exit code=${c} sig=${sg}`); });
setTimeout(() => {
  console.log(`after 4s: s0 exited=${exited.s0}, s1 exited=${exited.s1}, captured ${out.split('\n').filter(Boolean).length} lines`);
  console.log(exited.s0 ? 'OK: upstream exited' : 'HANG: upstream still running after downstream exited');
  s0.kill('SIGKILL'); s1.kill('SIGKILL'); process.exit(0);
}, 4000);
```

**Output:**
```
s1 (head) exit code=0 sig=null
after 4s: s0 exited=false, s1 exited=true, captured 5 lines
HANG: upstream still running after downstream exited
```

**Real test to pin it against actual ExecV3** (yours to write, since I can't run the package): an ExecV3 run of `yes | head -n5` (or `cat /dev/zero | head -c10`) — assert it **returns promptly** with head's output, not after `input.timeout`. If it only returns on timeout, the bridge needs to tear the upstream down when the consuming stage resolves.

bananabot9000:
## Re-verified against current head (I'd been on a stale checkout)

I'd read `afa75754`; the actual head is `726258dd`. The diff between them for `runPipeline.ts` is **only** the redirect-path fix (`resolve(cwd, redirect.stdout/stderr)`) plus formatting — the bridge / early-exit structure is unchanged. So:
- The **early-exit hang finding stands** against current code.
- Bonus: the **redirect-cwd bug I flagged earlier is now fixed** by that `resolve(cwd, …)` — good.

bananabot9000 (approved):
Approved — reviewed against the actual current head `726258dd`.

- The BLOCKED-program path is now a `ToolRefusedError` throw (→ `refused` outcome), not a fabricated results array — cleaner, and it moots the earlier BLOCKED-results-shape thread.
- Redirect paths now `resolve(cwd, …)` — the redirect-cwd issue I flagged earlier is fixed.

**One open item, non-blocking (on record, not a gate):** the pipe early-exit hypothesis is *unconfirmed*. `runPipeline`'s bridge structure is unchanged, and a standalone `spawn`+`PassThrough` mimic shows a downstream early-exit doesn't tear the upstream down — but I could not run actual ExecV3 (no deps), so this is a **worth-checking, not a confirmed bug**, and it's bounded by `timeout` regardless. A `yes | head` test against real ExecV3 (assert it returns promptly, not on timeout) settles it either way.

PR #381, opened 2026-06-30
Redesign composable pipe tools into atomic stages over typed streams

shellicar:
## Summary

- Split composable pipe tools so none is both a source and a stage; each takes its own input over typed streams
- Replace Grep and SearchFiles with Match and Paths | Read; ReadFile stays the non-pipe way to read
- Match takes grep-style before/after context, and a pipe skips binary files instead of failing
- File discovery carries type, size, and symlink target per record, making each path a first-class field
- Fix pipe stages being silently auto-denied, and report an unknown tool as a lookup failure rather than a false user rejection
- Validate regex pattern fields as schema errors before any tool runs


bananabot9000 (commented):
## Triage: fix-now vs needs-more-thought

Not gating this on the streaming/perf stuff — that's a redesign, not a blocker. The design itself holds: three-face composable + `reconcile` as the single graft + the runtime pre-flight is sound.

### Fix now (small, localized)
- **`Read` binary handling → skip, don't abort.** `Read.run` `throw`s `PipeStepError` on a binary file, which is fatal to the *whole* pipe — while a directory in the same stream is silently skipped (`continue`). So one PNG swept up by `Find | Read` kills a content search over every text file that read clean. grep's long-standing behaviour is note-and-continue (exit 0, keeps going); match it — skip the binary like you skip a dir. Optionally emit a skipped-list so the model isn't blind. One localized change in `Read.run`. This is the only thing I'd want in before merge.

### Minor / optional (not a blocker)
- **`permissionTools` carries live-handler stage tools.** `permissionTools = [...tools, ...stages.map(toStandalone)]` — those `toStandalone(stage)` objects have handlers that would `reconcile(model, undefined)` → crash on `input.kind` if ever invoked. Safe today only because the list is read for `.name`/`.operation`, never run. By-convention, not by-construction — a comment or a name-only shape locks it.

### Needs more thought (defer — explicitly not blockers)
- **Batch, not stream** (the "find | grep is slow" one). `Stream = FileRecord[]` is materialized; each stage runs to completion before the next sees a byte. `Find | Read | Head 30` reads the whole tree to keep 30 lines. Async-iterable streaming buys `Head` early-exit + O(1) `Read` memory — transport change, abstraction survives, but it's a redesign.
- **Validation placement.** Pipe centrally pre-flights the chain. The alternative — a handshake per joint over the stream — makes each tool self-sufficient and composable beyond the Pipe (Exec included). Architectural; separate conversation.

**Bottom line:** merge-worthy once the `Read` skip is in; everything else is either optional or a deliberate follow-up.

bananabot9000 (commented):
## `Read` still fatal on binary — the skip-vs-fatal item is unchanged

Confirmed against this update: `Read.run` still does `throw new PipeStepError(...)` when `fileTypeFromBuffer` sniffs a binary, while a directory in the same stream is silently `continue`d. So the asymmetry stands — a single binary swept up by `Find | Read` (a png in `docs/`, a font, a fixture) still aborts the whole content read over every text file that read clean.

We landed earlier on **uniform skip** — grep's note-and-continue (verified: `binary file matches`, exit 0, keeps going; never fatal for a discovery walk). Flagging that it isn't applied here. Fatal still fits `Paths { ['x.png'] } | Read` (you named it), but not a `Find`-driven sweep.

Not re-blocking on it — just confirming it's deferred, not done. Same for the `readFile` catch → fatal a couple lines up.

bananabot9000:
## Correction — retracting the fatal-on-binary note (I reviewed stale code)

My earlier "Read still fatal on binary" note is **wrong** and I'm retracting it.

Root cause on my end: my local review branch was stuck at `ea69cb37` (my first-review commit from hours ago). My re-fetch was silently *refused* because that branch was checked out, so I read old `Read.ts` while quoting the current head hash — I conflated `gh`'s remote head with my stale local content.

At the **actual current head `8b0a8471`**, `Read.ts` **skips binary**:
```
if (sniff) {
  continue; // binary file: no text lines to contribute, so drop it (grep -I). Read binary with ReadFile, outside a pipe.
}
```
That's exactly the uniform-skip (grep `-I`) we landed on — and it matches a real `Find | Read` over mixed pdf + text: text comes through, the pdf is dropped. Nothing to fix here; sorry for the noise.

bananabot9000 (approved):
Approved — reviewed against the actual current head `8b0a8471` this time (not the stale `ea69cb37` my checkout was stuck on).

- `Read` skips binary (`continue // grep -I`) — the uniform-skip we landed on. My earlier "still fatal" note reviewed stale code; retracted.
- Pipe pre-flight (source-first, edge-compat, per-step model parse) + run-loop (reconcile → run, catch → fatal) are sound.

Scope, honest: I verified the core (Read, Pipe validation/run) at the current head, not all 74 files — but the concern that mattered is resolved and you ran it live.

PR #382, opened 2026-06-30
Release 1.0.0-beta.13

shellicar:
## Summary

- Bump every published package to 1.0.0-beta.13


bananabot9000 (approved):
Release mechanics verified: 6/6 published packages bumped to `1.0.0-beta.13` consistently, CI green. Changelog regen is a confirmed no-op — the pipeline blocks when changelogs would change, so green CI is the proof they're already current (convention enforced structurally, not by trust). Clean release. 🍌

PR #383, opened 2026-06-30
Bump platform package to beta.13, name launcher process

shellicar:
## Summary

The beta.13 release missed the per-platform binary package. Its version stayed at beta.12, so the CLI's publish workflow failed on a version mismatch and claude-sdk-cli never published. This bumps the platform package to beta.13 so the CLI release can publish, names the launcher process so it can be matched by name, and documents the rule so the platform package is not missed in future releases.

## Changes

- Bump @shellicar/claude-sdk-cli-darwin-arm64 to 1.0.0-beta.13 so the CLI release publishes
- Set the launcher process title to claude-sdk-cli, so the launcher process is findable by name alongside the SEA binary it runs (the binary already sets the same title from main.ts)
- Document the platform package in CLAUDE.md and the rule that every platform package bumps with claude-sdk-cli; also add exec-core and the platform package to the Packages table

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,6 +38,8 @@ Full detail: `.claude/five-banana-pillars.md`
 | `packages/claude-sdk-tools/` | Tool definitions: Find, ReadFile, Grep, Head, Tail, Range, SearchFiles, Pipe, EditFile, PreviewEdit, CreateFile, DeleteFile, DeleteDirectory, Exec, Ref, TsDiagnostics, TsHover, TsDefinition, TsReferences |
 | `packages/claude-core/` | Shared: IFileSystem, expandPath, ANSI/terminal utilities |
 | `packages/mcp-exec/` | MCP server wrapping Exec tool |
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Verified the fix is complete:
- `claude-sdk-cli-darwin-arm64` bumped beta.12→beta.13, and it's the *only* platform package — nothing else stranded at beta.12, so this doesn't repeat the gap it's fixing.
- `claude-sdk-cli` is beta.13; optionalDependency pins `workspace:*` → resolves to beta.13 at publish. Versions line up, publish-platforms job will pass.
- `launcher.mjs` `process.title = 'claude-sdk-cli'` is clean and well-commented; the SEA child already sets the same title from main.ts, so both processes are findable under one name.
- CHANGELOG.md + changes.jsonl carry the launcher entry consistently (drift-gate green).

The CLAUDE.md rule (walk `platforms/*` alongside `apps/*`/`packages/*`, enforced by publish-platforms failing on mismatch) is the right CS-resistant fix — the workflow catches it, not memory. Clean. 🍌

PR #384, opened 2026-07-01
Executor-owned pipeline teardown (starting point)

bananabot9000:
## New Executor: pipeline-owning teardown — **draft / starting point**

Not finished, not tested — a starting point for you to build on.
- **Unsigned** — my session's `gpg-agent` is dead; please re-sign.
- **Not type-checked or run** against the package (no `node_modules` in my sandbox). Treat this as the mechanism sketched in the real shapes, not a tested implementation.

### Problem
`run()` is single-command, so ExecV3's `runPipeline` bolts the pipe lifecycle onto a primitive that never sees its sibling stages. Nothing tears the upstream down when a downstream stage exits — so `yes | head` / `find / | head` / `cat big | grep -m1` don't short-circuit; the upstream runs until the abort/timeout fires.

### Change
A `runPipeline(commands, opts)` on the Executor — the thing that spawns the stages owns their inter-stage lifecycle. Mechanism: **when a stage settles, destroy its feeding bridge and group-kill the upstream** (its consumer is gone); propagates upstream one hop per `close`. The SIGPIPE analogue a single-command primitive structurally can't do.

### Verified vs not
- **Verified in isolation:** the teardown, via a standalone `spawn`+`PassThrough` repro — `yes | head` resolves in **~66ms** vs hanging to the timeout (repro on #380).
- **NOT verified:** this exact code — no type-check, no run.

### Rough edges (yours to settle)
- Exit status of a torn-down upstream: `SIGTERM` here vs a synthesised `141`/SIGPIPE.
- Sink-flush ordering: `run()` awaits `finished`; this ends once.
- Error/redirect semantics; `run()` as a length-1 pipeline vs coexisting.

### DIFF
--- a/packages/exec-core/src/Executor.ts
+++ b/packages/exec-core/src/Executor.ts
@@ -1,6 +1,6 @@
 import { spawn } from 'node:child_process';
 import { existsSync } from 'node:fs';
-import type { Writable } from 'node:stream';
+import { PassThrough, type Writable } from 'node:stream';
 import { finished } from 'node:stream/promises';
 import type { CommandSpec, ExitStatus, IExecutor, SpawnOpts } from './types.js';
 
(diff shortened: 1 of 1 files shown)


bananabot9000:
## Self-review — real bug in the aborted-at-entry guard (flagged by @shellicar)

Traced against `run()`. The `addEventListener('abort', …)` is fine on its own — the top guard (`if (opts.signal?.aborted) return …`) catches the already-aborted case before the listener is reached, so the classic never-fires staple is covered.

**But the guard is an incomplete copy of `run()`'s and it hangs the caller:**
```ts
// run()'s guard:
if (opts.signal?.aborted) { await closeSinks(opts); return { exitCode: null, signal: 'SIGTERM' }; }
// runPipeline's guard (mine):
if (opts.signal?.aborted) { return commands.map(() => ({ exitCode: null, signal: 'SIGTERM' })); }
```
Mine returns **without `closeSinks(opts)`**, so an already-aborted pipeline leaves the caller's `stdout`/`stderr` open → whoever drains them (a `fromStream` capture) never sees `end` → hang. Same hang class the PR is meant to *fix*, reintroduced in the guard.

Fix: `await closeSinks(opts)` (de-duped) in that branch before returning, like `run()`. One-liner.

bananabot9000:
Superseded by #402, which is this design productionised — teardown owned by the pipe layer (ExecV3), exec-core providing the reason-mapped kill, real SIGPIPE deaths, and tests against the actual package (the exact gap this draft admitted in its own STARTING POINT comment). The mechanism this draft proved in isolation lives on in there, grown up.

PR #385, opened 2026-07-01
Render assistant responses as styled markdown in the terminal

shellicar:
## Summary

- Parse assistant responses with `marked` and render tokens to styled ANSI
- Highlight fenced code with cli-highlight in a boxed, language-labelled block
- Cap and wrap long code lines so they stay on screen instead of clipping
- Two opt-out config flags (render markdown, render during streaming), both default on


bananabot9000 (approved):
## Reviewed the render core + the streaming-partial risk

The risk in a markdown renderer over *streaming* text is running `marked.lexer` on incomplete syntax (half-open code fence, unterminated `**`) — reflow, dangling ANSI, height jumps fighting the count-preserving contract. Checked it:

**Correctly gated (verified).** Two flags: sealed `response` blocks render markdown via `renderBlockContentCached` (computed once, cached); the *live partial* block only renders markdown behind a separate `markdown.streaming` flag. So the risky path is opt-in, not the default. When it's on: incomplete inline (`**bol`) stays *literal* in `marked` until closed → no dangling ANSI escape; partial code stays count-preserving via `getHighlighted` (line-preserving on every path); the per-frame repaint absorbs the height changes. Sound design — the author saw the partial-render risk and fenced it off.

**One real minor finding (non-blocking, cosmetic).** In `list()` → `emitLines(pad + marker + inline(text), cols)`: `inline()` turns a `br` into `\n`, and `emitLines` splits on `\n` *after* prepending `pad + marker`. So only the first line gets the indent+marker — a list item with a hard line break renders its continuation line flush-left, no indent:
```
• first line
second line        <- lost the two-space indent
```
Rare input (a `<br>` inside a bullet), purely visual. Fix: split first, then prefix only line 0 and indent the rest.

**Coverage:** read-verified (the streaming gate + the token walker), not run-verified — no `node_modules` in my clone, so I didn't execute the count-preserving invariant against the measurement path or fuzz constructs. Tables/task-lists fall through to raw passthrough by design (fine).

No blocker.

bananabot9000:
### Correction on the list-item / `<br>` note — downgrading it

I called that a "real minor bug." It isn't confirmed — I couldn't run `marked` (no deps in my clone), so I've only verified the *downstream* half by reading: `inline()` maps a `br` → `\n`, and `emitLines(pad + marker + text)` prefixes then splits on `\n`, so a continuation line loses its indent. That part holds.

What I did **not** verify: that a list item's inline tokens ever actually contain a `br`. Candidate trigger is a hard break in a bullet (`- first␣␣\n  second`), but whether `marked` emits a `br` there vs a soft break, I don't know without running it.

So: **candidate, not a confirmed finding.** One-liner to settle it — `marked.lexer('- a  \n  b')`, look for `type: 'br'` in the item's tokens. If it's a soft break, disregard.

PR #386, opened 2026-07-01
Survive a mid-turn network drop instead of losing the turn

shellicar:
## Summary

- Keep the machine awake for the duration of a request so sleep cannot drop the connection
- Retry a mid-stream connection drop on a bounded schedule instead of failing the turn
- Persist the conversation as each message is sent and answered, so an interruption cannot lose history
- Resume an interrupted turn from an empty submit


bananabot9000 (approved):
## Reviewed the resilience path — the two places it could break are both closed

I went at the network-drop path where the bug hides, not the happy case.

**Retry can't inherit the dead stream's state (verified).** On `StreamInterruptedError` the loop `continue`s → `processor.process(stream)` again, and `StreamProcessor` is a reused singleton — so the question is whether a second `process()` inherits the interrupted first stream's partial blocks. It doesn't: `currentToolId`, `hasToolBatch`, and `new MessageAccumulator()` are all **locals** inside `process()`, fresh per call. The reconnect assembles a clean result. (And `StreamProcessor.ts` is untouched by this PR — no regression.)

**No double-render on reconnect (verified).** `process()` throws mid-iteration *after* emitting `message_start`/`message_text` but *before* the closing events, so the subscriber is left with a partial. `StreamInterruptNotice.reconnecting()` handles it: `completeActive()` seals the partial as a finished block ("the next attempt can't append onto it"), then splices the notice. The retry's re-stream starts a fresh block below it — no doubled text. And only the *complete* stream-2 result is pushed to conversation history, so the partial is on-screen residue only, never in the session. Correct.

**Non-blocking notes:**
- `STREAM_INTERRUPT_DELAY_MS = 30_000` is a long *fixed* wait for a case that recovers on network-return time — a socket back in 2s still waits 30s. You've flagged it tunable ("SC's starting points"), so just naming it.
- **Coverage gap I'll own:** I didn't trace interrupt *mid-tool-batch* — a stream that emitted `tool_batch_start` but died before `tool_batch_end`. Does `completeActive()` seal a tools-block-in-progress as cleanly as a text block? Worth a test; I didn't verify it.

Reaches the bar. No blocker found.

PR #394, opened 2026-07-04
Show user, tools, and claude time totals in the status line

shellicar:
## Summary

- Add a per-side time breakdown to the status line: user, tools, and claude totals
- Accumulate time through a pure event clock that listens to six role edges
- Count only successful (2xx) request time as claude; failed and retried time stays untracked
- Repaint the status line while idle so totals stay current, and lock the render with deterministic tests


bananabot9000 (commented):
## Verified the clock edges against the state machine's own contract — one real leak

Read `TurnClock`, both adapters, the TurnRunner diff, QueryRunner's `#handleTools`, EditorHandler, and the main.ts wiring, at head `31444bd1`.

**The core is clean.** Stop-no-ops-on-mismatched-role is an elegant way to express deferred attribution, and `claudeStop(kept)` is wired correctly through the whole retry loop: every attempt is a closed bracket — `requestStarted()` at the top of each iteration, `requestSettled(true)` on success, `settled(false)` in the catch *before* every continue/throw (abort, account-limit, stream-interrupt, transient). Backoff sleeps sit between brackets → untracked Unknown, exactly as documented. User bracket (editor-open → submit) closes on both submit paths. The unref'd 1s repaint is right.

**The leak — multi-tool batches charge approval waits to `tools`, contradicting the code's own comment.** `#handleTools` says: *"one bracket from the first local execution to the last. Approval waits … sit outside it."* That holds only for the wait before the **first** execution. With `requireApproval`, the loop is: approve A → `beginToolExecution()` (tools clock **starts**) → run A → back to `await Promise.race(pending)` for B's approval — **inside the open bracket**. `toolsStopped()` only fires after the last tool.

Repro shape: batch of two Execs. Approve A instantly (runs 1s), leave B at the prompt for 5 minutes, approve (runs 1s) → `tools` total ≈ 5m02s, actual tool execution 2s. Sequential one-by-one approvals are the normal usage pattern, so this hits the primary stat the feature exists to report.

**Fix that drops out of your own design:** per-run brackets — `toolsStarted()` before each `run`, `toolsStopped()` after (the stop-no-op semantics make repeated brackets safe by construction, and it makes the comment true). Cosmetic side effect: the active-role highlight flicks off between tools of a batch — arguably more honest anyway.

Two half-line notes, not findings: (1) if a `run()` ever throws past its `finally`, the trailing `toolsStopped()` is skipped — the state machine self-heals (segment lands in Unknown, no corruption), just undercounts. (2) The user bracket has no wake-lock (correctly), so lid-closed-overnight-mid-edit lands in `user` — chess-clock-honest, but worth knowing it's in the stat.

Not reviewed: `formatDuration` internals and the test files beyond skim. Holding the stamp for your call on the batch-wait leak: fix in this PR or note-and-ship.

bananabot9000 (approved):
Approved. The batch-wait behaviour I flagged is the intent — it's a chess clock: idle time must land somewhere, and stopping the timer while an approval sits at the prompt would make the clock under-report, the one thing it exists not to do. My proposed per-run brackets were backwards.

What I actually found was a contradiction between the code and its comment (`Approval waits ... sit outside it (not tools running)` — true only before the first execution, not between tools in a batch), and I decided the comment was the spec instead of asking which one carried the intent. The code is right; the comment describes a different clock. Non-gating — leave it or touch it up in passing.

Everything else verified and clean: `claudeStop(kept)` correct through all four retry paths (abort, account-limit, stream-interrupt, transient), every attempt a closed bracket, user edges close on both submit paths, unref'd 1s repaint.

bananabot9000:
Correcting my close-out: the comment is a known defect, not a take-it-or-leave-it — it will mislead the next reader exactly the way it misled me. Fix it before merge. Suggested replacement for the two lines in `#handleTools`:

```ts
// Tools clock: one bracket from the first local execution to the last tool.
// The wait before the first execution is untracked; once the bracket is open
// it runs through approval waits between tools — chess-clock semantics, idle
// time lands on whichever clock is open. Resolve-only errors never start it.
```

Approval already stands, so this won't block the merge mechanically — but the ask is: fix the comment, then merge.

PR #395, opened 2026-07-05
Add --claudeMd flag and cache the assembled CLAUDE.md prefix

shellicar:
## Summary

- Add a `--claudeMd` flag that contributes a string to the assembled CLAUDE.md content at launch
- Serve the assembled CLAUDE.md prefix from cache on repeat launches instead of paying for it each turn
- Guard the prefix cache marker so it only marks a genuine system-reminder boundary block
- Keep the CLAUDE.md context present in every request after compaction, where it previously dropped out and never returned
- Document the request cache-breakpoint budget in the repo CLAUDE.md


bananabot9000 (commented):
Verified at a9fc5bea — traced the full path: `--claudeMd` → `IRuntimeOptions` → loader section → `DurableConfigFactory.update()` per turn → `QueryRunner` first-message injection → `cacheClaudeMdPrefix` pinned breakpoint. The design is sound and the budget doc is the honest kind (4/4 spent with CLAUDE.md, no headroom, fifth rejected — named as a live ceiling). Flag plumbing mirrors `--system` cleanly. The guard's cross-run drift cases check out safe: count mismatch in either direction degrades to a no-op or a one-time cache miss — cost, never corruption.

Three things before I stamp, first one's the real question:

**1. Post-compaction re-injection: the comment and the code disagree — which carries the intent?** The comment says the injection covers "a post-compaction state where the original first user message … has been dropped." The code requires something stronger: `!messages.some(m => m.role === 'user')` — *no user messages at all*. If your compact path leaves any user message in history (e.g. recent turns survive, only old content is cleared), the condition never fires: the original reminder blocks are gone, re-injection never happens, and CLAUDE.md silently drops out of context for the rest of the conversation — while the pinned-marker call no-ops on the guard, so nothing errors. If compaction truly empties all user messages, the code is right and the comment just understates the condition. Which does your compact output actually look like?

**2. Mid-session CLAUDE.md edits are now conversation-pinned — confirming it's the intended trade.** The loader re-reads files every turn, but the model sees the copy baked into message 1; current content only supplies the *count*. Edits land at the next fresh conversation or compaction re-injection. That's the right trade for cache stability (and compaction becoming the refresh point is a nice bounded-staleness property) — but the loader's docstring still promises "changes are picked up without any watcher," which is now true of the read and not of what the model sees. One clause would square it.

**3. Half-line, the old scar:** when compaction runs with all 4 breakpoints spent, does the compact path itself add any marker? Asking because a fifth was exactly the cache_control bug from the April incident — presumably checked, just want it said out loud.

bananabot9000 (approved):
Approved.

Correcting my review: the "conversation-pinned **now**" finding was wrong — QueryRunner is untouched by this PR (0 lines), the first-message injection predates it. I read the injection code at the PR head and folded pre-existing behaviour into my model of the change, when the changed-files list was telling me otherwise. The PR doesn't alter when the model sees CLAUDE.md at all; it only stops re-billing the already-pinned block. Pure billing win, no behaviour change. The loader-docstring nit is likewise pre-existing, not a PR finding.

Compaction questions: moot per your usage (compact disabled) — the comment/code difference in the injection condition is in code this PR doesn't touch anyway.

What this PR actually is, verified: --claudeMd flag plumbed cleanly (mirrors --system), a pinned cache breakpoint with a shape guard whose failure modes degrade to cost-never-corruption, and an honest 4/4 budget doc. Clean.

bananabot9000:
Follow-up on the compact question ("does this fix CLAUDE.md loss on compact?") — checked against the code, and the answer is **no, and it can't, and the reason also resolves my earlier comment/code question**: the comment was describing a state this architecture never produces.

Three code-verified facts:

1. **CLAUDE.md rides only in the first user message.** Injected once, when history contains no user message; after that, today's content only supplies the breakpoint count.
2. **Compaction removes it from the wire.** `Conversation` keeps full history forever ("compaction messages… do NOT cause prior history to be dropped"), but `cloneForRequest()` trims every request to the slice *from the last compaction block forward* (`trimToLastCompaction`). The first user message — and the CLAUDE.md blocks inside it — fall off every post-compaction request.
3. **The re-injection can never fire.** The condition tests `this.conversation.messages` — the **full** history, which still contains that first user message forever. So `some(role === 'user')` is true for the life of the conversation, including after every compaction. The QueryRunner comment ("covers… a post-compaction state where the original first user message has been dropped by the API") presumes an API that drops history; this Conversation never lets that happen. The re-injection branch is dead code for compaction purposes.

So your hypothesis holds with a mechanism attached: **after a compact, the session loses its entire constitution** — every convention and rule in CLAUDE.md stops being sent — which would read exactly as "compact is horrible": the summary wasn't the (only) problem, the standing orders vanished with message 1.

**Where the fix belongs: RequestBuilder, not QueryRunner.** It already owns request-time mutation and already knows `cachedReminders` (content *and* count). Make CLAUDE.md a request-window invariant instead of a message-1 artifact: when the first user message of the request slice doesn't open with the reminder run, prepend the blocks there (request-scoped mutation on the caller-owned clone — same ownership contract the ephemeral reminders already use, and the pinned cache marker is already in the same file pointing at the same position). That fixes every compaction, keeps history unmutated, and turns the stale QueryRunner comment true by deletion — the injection there stays purely the fresh-conversation path.

bananabot9000:
Correction: the analysis above was read at a9fc5bea — stale; the fix had already been pushed. Verified at 29b7f031: `ensureClaudeMdReminders` (TurnRunner, on the caller-owned clone after `cloneForRequest`) re-injects the reminder run when the post-compaction slice's first user message doesn't lead with one, and no-ops pre-compaction — so CLAUDE.md is a request-window invariant now. Checked the interplay: `cacheClaudeMdPrefix` keys off the same shared block shape (`buildReminderBlocks` as single source), so the pinned marker lands correctly post-compact; ephemeral reminders are injected after `ensure` and never persist, so they can't fool the idempotence guard. The new TurnRunner test covers the exact scenario. Fixes the issue. Approval stands.

PR #396, opened 2026-07-05
Surface the API error detail on a failed request, not just the status

### LINKED ISSUE #392: HTTP error responses show only the status code, not the message
When a request fails, the CLI shows only the HTTP status, `[error: HTTP 400]`, and nothing else. The response body carries a message explaining what went wrong, and that message is dropped. For a 400 in particular this leaves you with no idea what was actually wrong with the request.

**Reproduce:** send a request that the API rejects with a 400. The CLI prints `[error: HTTP 400]` with no message.

shellicar:
## Summary

- Carry structured API error detail (status, type, message) across the SDK to CLI channel
- Bring non-HTTP stream error detail across the same way
- Format the error in the CLI so a failed request shows what went wrong, not a bare HTTP status

Closes #392

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -95,6 +95,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Prevent crashes from unhandled child process and socket errors
 - Reject unknown flags at launch instead of silently ignoring them
 - Restore cursor visibility after exiting the CLI (#277)
+- Show the API error detail on a failed request instead of only the HTTP status
 - Show the permissions notice only when displayed permissions change, not on every config edit
 - Stop duplicated content (ghost text) stranding at the wrap boundary in the TUI: the renderer now builds and diffs a cell grid and writes every row at an absolute position with autowrap disabled
 - Stop the CLI freezing on an account-limit retry-after wait; retries are capped, ESC-abortable, and give up with a single account-limit notice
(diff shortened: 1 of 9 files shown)


bananabot9000 (commented):
Verified at 1225e383 — traced the chain end to end: `HttpError`/`ApiStreamError` body → `readBodyError` (every step of the unknown-shape read is guarded, no unguarded casts) → `toSdkErrorDetail` → `detail` on the `SdkError` event (`...(detail ? { detail } : {})`, no explicit-undefined key) → `formatSdkError` in the CLI. Six SDK-side tests cover status/type/message/fallback/mid-stream/non-transport. Clean chain, right seam placement (extraction in the SDK, formatting in the CLI).

One real rendering wart, in exactly the area this PR exists to polish: **a body-less HTTP error renders its status twice.** `toSdkErrorDetail` falls back to `message: "HTTP 500"` when the body carries no message; `formatSdkError` then builds `prefix = "HTTP 500"` from the status and renders `${prefix}: ${detail.message}` → **"[error: HTTP 500: HTTP 500]"**. The 500-with-empty-body case is tested SDK-side (message === "HTTP 500") but the CLI composition isn't, so the duplicate slipped between the two layers. It's also a likely case in the wild — proxies and gateways love returning HTML bodies for 429/502/503, all of which fail `readBodyError` and take this path.

One-conditional fix in `formatSdkError`: skip the message when it equals the prefix (or when the detail carried no body message, render the prefix alone). Holding the stamp for your call — fix here or ship-and-shrug; it's cosmetic, but cosmetic is this PR's whole genre.

bananabot9000 (approved):
Approved. Ruling taken: the duplicate render is accepted, and the severity argument I attached was mispriced — 'proxies love HTML bodies' prices the generic internet, and this transport talks to exactly one API that returns proper JSON error bodies. Severity lives in the actual deployment, not the hypothetical one. The chain itself verified clean: guarded reads, right seams, six tests.

PR #397, opened 2026-07-05
Recover conversations by directory from a central session store

shellicar:
## Summary

- Record each session's directory and conversation id to a central sqlite store as the CLI runs
- Resume the most-recent conversation for the current directory instead of reading a per-directory marker file
- Remove the .sdk-conversation-id and .sdk-conversation-history marker files; the store is the single authority
- Reorder the store open pragmas so a concurrent startup herd waits for the lock instead of throwing


bananabot9000 (commented):
Verified at 4099aeb2 — and this one I reviewed twice in a sense, since I supervised the mission's agreement/provenance on Saturday. The requirement that mattered most there landed intact and is now *pinned by a test with the reasoning in its name*: "keeps both records when the same cwd is appended twice (a log, not an upsert)". The append-only decision — kept so lifecycle events can reconstruct active state later — survived Investigator → Builder → code → test without perversion. That's the fleet working.

Craft verified, not asserted:
- **Pragma order is right and tested on disk.** busy_timeout(5000) before synchronous/journal_mode, so the fresh-file delete→WAL switch waits instead of throwing under a startup herd — and DatabaseFactory.spec runs against a real tmpdir file (`inMemory: false`), asserting WAL-under-held-lock, journal_mode, and busy_timeout. The in-memory-fake-hides-the-prod-bug failure mode from the SqliteObjectStore saga is explicitly not repeated.
- **ORDER BY id DESC** — resume keyed on the autoincrement, not the timestamp. Monotonic under clock skew; correct choice.
- Migrations via user_version with the append-never-edit discipline; eager init in the container factory so open errors surface at boot; per-name memoised connections.

One real question before the stamp, and it's the upgrade story: **existing directories lose auto-resume on first launch of this build.** load() consults only the store; the store starts empty; `.claude/.sdk-conversation-id` is no longer read — so every directory with a live marker gets a fresh UUID, and the old conversation is stranded (recoverable by hand: the id is still in the dead marker, `--resume` works). The mission said "replace the marker," so the Builder built to spec — this question is for the intent-holder: deliberate one-time break, or worth the three-line seed (store-miss + marker-exists → append the marker's id before resolving)? Fleet-wide that's the difference between one odd morning and nobody noticing.

One carried-forward note, not a finding: two CLIs launched in the same cwd still auto-resume the same conversation (parity with the marker world; liveness was explicitly deferred, and the log leaves the room for it). Nothing else — the store side is clean.

bananabot9000 (approved):
Approved. Ruling taken on the upgrade break: deliberate clean removal, --resume covers the stragglers.

And on the first-verified-mission irony, for the record: provenance is a fidelity instrument, not a truth instrument. The chain proved you got exactly what the agreement said — which is only the guarantee you want when the agreement is right. My supervision ran under 'assume the intent is correct' by instruction; the one lens excluded was the one holding the bug. The store side of this PR is clean regardless — the flaw lived in the restore story, not the log.

PR #398, opened 2026-07-05
Split CI: build each package once on Linux, fan out native SEA wrap

### LINKED ISSUE #390: Build the Node app once and reuse it across platform builds and publish
The CI pipelines carry more maintenance burden than they should, and the shape is the reason.

The CLI ships as a per-platform SEA binary. In the publish workflow, each platform leg builds the workspace and then bundles its binary. The workspace build is the same on every platform, so it should be built once, as the output of a single job, and the platform legs should fan out from that one artifact to produce their binaries. Today only one platform is live, so the duplication is hidden; the moment a second platform is added, the same bundle gets built again per platform.

There are two workflow files: a checks gate that runs on pull requests, and a publish that runs on release. The publish repeats the build-and-package boilerplate across its jobs instead of sharing it, and the binary build only happens at release time, so a pull request never exercises the packaging path. I'd rather one pipeline that behaves by trigger: a pull request builds and packages, main optionally does the same, and a tag or release builds, packages, and publishes.

Part of the current pain is that the hidden GitVersion step does not work, and I suspect the pipeline cannot cleanly tell whether a tag, a commit, or a pull request triggered it. Reshaping around a single triggered pipeline should address most of this.

The goal is to reduce the maintenance burden of the CI pipelines, not to change what they produce.

## Guidance

The behaviour I'm picturing, by trigger:

- pull request → build & package
- main (optional) → build & package
- tag / release → build & package & publish

In scope alongside the two workflow files: the supporting build scripts, chiefly `apps/claude-sdk-cli/build-sea.sh` (the script that builds the workspace and then the SEA binary), and the local composite actions (`install-gitversion`, `publish-package`). This wants an investigation of the current pipeline before any restructure.

shellicar:
## Summary

- Build each package once on Linux, where GitVersion and the tooling work; this fixes the broken versioning with nothing changed in GitVersion
- Per-platform SEA legs wrap and sign the prebuilt bundle instead of rebuilding it
- Four reusable workflows (checks, build-package, publish-generic, publish-sea) plus a new setup composite
- One ci.yml with three triggers (PR, main, tag/release), with publish fenced to tag/release alone
- Mark mcp-exec publishable so the PR packaging filter includes it

## Delivery Notes

- Checkout stays a one-line first step in each job rather than moving into the setup composite. A local composite action only exists after the repo is checked out, so it cannot perform the checkout that makes itself readable. setup centralises pnpm and node; checkout is the one line that cannot move.

Closes #390


bananabot9000 (approved):
Verified at f04f7402, and the strongest evidence is this PR's own run: detect → three affected builds → **the wrap leg actually executed on macos-14 with the Linux-built blob** (1m05s, publish fenced off, launcher correctly skipped) — the whole build-once/fan-out seam exercised live before merge. The gitversion fix is structural, not configured: version work now exists only on the Linux leg; the macOS runner never invokes gitversion at all, so the osx-arm64 `error 13` class is unreachable rather than patched.

**Omission hunt (the deleted workflows' duties, traced one by one):**
- checks.yml carries every old node.js.yml step verbatim — build/type-check/test/`pnpm run ci`/scripts validate/changelog-drift.
- The old npm-publish changelog gate moved into publish-generic (base-version match + dated-entry check, intact).
- Version-verify → dist-tag → publish lives in the shared publish-package composite, with the #368 GITHUB_OUTPUT-pollution fix preserved (the capture-then-filter comment survived the move).
- The launcher job publishes without restoring the artifact — checked, correct: its payload is `files: [bin, CHANGELOG, README]` and `bin/launcher.mjs` is checked in; no dist in the launcher.
- The publishable filter is `private === false` (explicit opt-in) — confirmed every intended package carries it; mcp-exec's flag being this PR's addition; typescript-config/scripts correctly excluded.
- Branch protection has no required-check names to orphan (checked via API — empty), so the job renames can't wedge future PRs.

**Three notes, none gating:**
1. build-package's `artifact` output description says "empty when the package is not publishable," but the code sets the name unconditionally — only the *upload* is conditional. Doc-vs-code; either make the output conditional or fix the sentence.
2. Inherited, worth knowing: `workflow_dispatch` checks out default-branch HEAD, not the input tag — a dispatch "test" builds main-HEAD under the tag's version whenever package.json still matches. Same shape as the old flow; the new build-once world just makes it slightly easier to publish a commit newer than the tag.
3. On PRs, `wrap` needs *all* matrix builds green — an unrelated package's build failure skips the wrap check. Fine (the PR is red anyway), just naming it.

**What this run cannot prove:** the publish:true path — platforms→launcher ordering against npm, OIDC provenance, version-verify against a real tag. The first release is that exam; a beta tag as the guinea pig and this pipeline's earned its stripes.

Approved — the notes above are observations, nothing rides on them.

PR #399, opened 2026-07-05
Condition attached images to a safe size on attach

shellicar:
## Summary

- Resize and normalise every attached image to a 2000px PNG long edge via sips, downscaling only, so an oversized image can no longer exceed the request-size limit and take down the conversation
- Cover both attach paths: clipboard paste and image ReadFile results, leaving non-image documents untouched
- Degrade to attach-as-is when sips is absent or fails on an image, never blocking the attachment or the turn
- Log each conditioning outcome to the debug log, so the silent passthrough is visible in the field

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -64,6 +64,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - List --file in --help output
 - Move source files into `model/`, `view/`, and `controller/` subdirectories; add biome.json boundary enforcement
 - Repaint every TUI row each frame, resilient to external grid mutation (e.g. tmux reflow)
+- Resize and normalise a pasted image before it is attached, so an oversized image can no longer exceed the request-size limit and take down the conversation
 - Rewrite the project documentation: what the CLI is, why you would use it, and how to install, configure, run, and extend it
 - Serve the assembled CLAUDE.md prefix from cache on repeat launches instead of paying for it each turn
 - Set the launcher process title to claude-sdk-cli so the launcher process can be matched by name, alongside the SEA binary it runs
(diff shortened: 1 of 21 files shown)


bananabot9000 (commented):
Verified at 5090ee3d (fetched fresh). The core is right and the contract is honest — findings below, one asymmetry worth your ruling.

**Verified sound:**
- `conditionImage` gate/downscale/degrade logic is clean: dimension-gated so `-Z` only ever runs on oversized images (downscale-only holds), PNG re-encode propagates the *new* media type through both attach paths — no bytes/type mismatch anywhere.
- `parseDimensions` throwing on unreadable output → degrade is the right polarity: an unparseable sips response attaches-as-is with a warn instead of silently mis-gating.
- `NodeSipsBridge`: `execFile` with an args vector (no shell surface), 10s timeout (timeout kills + rejects → degrade), `mkdtemp` per call (concurrency-safe), cleanup in `finally`. Bare `sips` on PATH is consistent with the existing pngpaste/osascript precedent and documented as such.
- ReadFile's 5MB cap check moved from pre- to post-conditioning — correct order: cap what would actually be sent, so a conditioned image passes and a degraded-but-oversized one still errors helpfully.
- The bridge abstraction is the product answer: the platform dependency is one injected seam, so a future conditioner (wasm, Rust sidecar, anything) drops in without touching either attach path. sips-as-engineering, not sips-as-workaround — the difference is the contract, the logging, and the swappable boundary.

**The one asymmetry — paste path has no byte backstop.** ReadFile degrades *safely*: if sips is absent/fails, the 5MB cap still gates the unconditioned image. The paste path degrades *all the way back to the original hazard*: sips absent/broken → original bytes attach uncapped → the exact oversized-request kill this PR exists to prevent, reachable again on any non-Mac or a broken sips. On your Macs sips is always there, so this may be accepted risk — but the harmonization is cheap (same post-conditioning size check before `addImage`, warn + skip). Your call: accepted residual or one more guard.

**Half-line:** an oversized *animated* GIF flattens to a static PNG on conditioning — content loss the model wouldn't have seen anyway (image blocks are static), so noting for completeness, not action.

Holding the stamp only for the paste-backstop ruling — everything else is clean, tested (gate/resize/degrade all covered, both wiring paths updated including mcp-exec's entry), and the field-visibility logging is exactly what the silent-passthrough case needs.

bananabot9000 (approved):
Approved. The paste-backstop finding was mispriced as a PR finding — the paste path never had a byte cap on main; this PR only adds conditioning in front of it, so the diff strictly improves that path and the degrade case merely returns to the status quo ante. Pre-existing gap, surfaced for the backlog, not a gate on this diff. Everything else as reviewed: gate/downscale/degrade verified, post-conditioning cap order in ReadFile correct, bridge seam clean.

PR #400, opened 2026-07-05
Bind a conversation's system identity so it survives resume

shellicar:
## Summary

- Add `--system-identity <file>` so a conversation owns the actor it casts Claude as, not the process
- Persist the identity file path in sqlite keyed by the conversation id; `--resume <id>` brings the actor back with no flag
- Read the file body live from disk every query as the first (base) system prompt, ahead of SYSTEM.md and --system
- Drive the status line from the frontmatter `name` ("unknown" when missing or malformed)
- Degrade to a warn notice when the file is absent at load; assertion on a missing file errors and does not launch


bananabot9000 (commented):
Verified at c70d53dd. The design is sound end to end — one blocking item (the red CI, root cause below), one on-the-record trade, one reuse note.

**The red CI, named:** `checks` fails at type-check — `test/QueryRunner.spec.ts(102,7): TS2515: FakeDurableConfigProvider does not implement updateIdentityBody`. The new abstract member landed on `IDurableConfigProvider`, the CLI concrete got it, the SDK's own test fake didn't. One no-op line in the fake. Note lint/test never ran (type-check failed first), so green still has to prove the rest of the gauntlet after the fix.

**Verified sound:**
- **The ownership model is the right one.** The conversation owns the actor; the process just casts it. `assert` = set-and-persist unconditional (flag), `load` = defer to what's owned (no flag), `inherit` wired into Ctrl-/ n. Path durable in the objects store keyed by conversation id; body and name never cached.
- **Live-mirror semantics are coherent at every layer**: body read fresh per query in runTurn (the only moment it reaches the model), display name refreshed by a file watch (correct asymmetry, and the comment says why), deleted → contributes nothing this turn + status segment drops, restored → back. Degrade-never-block polarity throughout, with the one deliberate fatal — assert-time existence — placed at the only moment a typo'd path should stop the world.
- **Prompt ordering as stated**: identity is the first base system block, ahead of SYSTEM.md and `--system`; empty/null contributes nothing.
- **parseSystemIdentity**: CRLF-tolerant anchored frontmatter, lazy body split, malformed YAML degrades to `name: null` without throwing (a bad identity file can never kill the prompt), frontmatter never reaches the model, "unknown" only when the file is present and nameless.
- Watch disposed in cleanup beside the config watch; no mid-session path acquisition exists, so the watch-only-at-launch gate is complete.

**On the record, not gating:** the identity body is the *first* system block and live-read — an edit mid-conversation invalidates the whole cached prefix (system, tools, and the CLAUDE.md marker chain behind it) for one turn. That's the live-mirror trade, chosen with eyes open; identity files change rarely. Stating it because #395 just spent a night making that prefix stable, and this deliberately parks the one volatile thing at the front of it.

**Reuse note:** the inline `~` expansion in main.ts (`replace(/^~(?=\/|$)/, HOME)`) duplicates `expandPath` from claude-core, which ReadFile already imports. Same behaviour, one source.

Holding the stamp for the green run — the fake is a one-liner, then this is clean.

bananabot9000:
Correction to my cache note: the segments are cached separately — the order is tools → system → messages, each breakpoint caching its own prefix independently. An identity edit therefore re-writes the system segment and the message-level segments downstream of it, but the tools segment (upstream of system) cache-reads untouched, and the CLAUDE.md *content* re-caches at the same position rather than being lost. Smaller blast radius than my "whole cached prefix" implied; the trade note stands at its corrected size.

bananabot9000:
Retracting the reuse note — I checked expandPath after asserting instead of before. It is NOT the same behaviour: it expands `$VAR`/`${VAR}` as well as `~`, so swapping it in would add env expansion to the identity path (a behaviour change, not a dedupe). And it takes an injected IFileSystem, which doesn't exist at arg-parse time — the container isn't built yet at that call site. Wrong twice in one sentence; the inline expansion is correct as written.

bananabot9000 (approved):
Approved at cee5b873. The three commits since my review are exactly the exit criteria: the TS2515 fake stubbed (one no-op line, as diagnosed), lint appeased, changelog housekeeping — and checks is green with the full gauntlet run this time (type-check, lint, tests all reached). Design verified in the prior pass; the cache note stands at its corrected size; the reuse note was retracted as wrong. Clean.

PR #401, opened 2026-07-05
Publish conversation activity as NATS tap events

shellicar:
## Summary

- Publish each conversation's activity as versioned events on `tap.v1.{conversationId}.events`, so running sessions are observable without window-hopping
- Disabled by default with zero effect; enabled fails fast on an unreachable broker at startup and tolerates mid-run disconnects without interrupting the conversation
- Add a JetStream compose stack, a subscribe script, and conformance tests against the versioned wire schema


bananabot9000 (commented):
Verified at 13e65e8b (re-fetched after a head race — the first fetch caught 898f64e1 mid-push). CI redness ignored per instruction. Stage 1, real: and the kit from tap-testing.md is in here as actual files — schema + crash-resume fixture validated as the artifacts, not restatements, driven through the real NatsTap with a capturing transport, fixed clock, and fake-timer heartbeat. The open-world schema is exemplary: `additionalProperties: true`, `type` a plain string (not an enum) with if/then per-type requirements — the closed-world trap from the tap-testing doc, dodged in the artifact itself. The transport's laziness comment is TRUE (dynamic `import('@nats-io/transport-node')`; the static import is type-only), fail-fast/tolerate lands exactly as the spec argued it (`waitOnFirstConnect` off for the startup gate, `maxReconnectAttempts: -1` for mid-run, publish-never-throws with the honest-hole drop), and `run_ended` is clean-exit-only with the 500ms bounded race in cleanup so a dead broker can't hold the process open.

Two findings, both with a live repro in your own 4:39 capture:

**1. `approval_settled` can publish before `approval_pending`.** Mechanism located: the pending event rides the sdkChannel subscription, the settled event rides the consumerChannel subscription — two channels, one subject, no cross-channel sequencing. On auto-approve the response lands first, and NATS faithfully preserves the inverted publish order. It matters because a consumer folding run state ends on `pending` — a phantom "waiting on you" that never clears until the next event. Fix direction is yours: a single emission queue for both projector paths, or emit pending at the request's consumer-side sighting so both come from one stream.

**2. Turn events are mismatched granularities.** `turn_started` maps `message_start` (fires per API request, each round of the loop); `turn_ended` maps `done` (fires per exchange). Your capture shows the result: two `turn_started`s, no `turn_ended` between. The spec's table implies pairs. Which does "turn" mean? If the exchange: only the first `message_start` of a query should project. If the request round: `turn_ended` needs a per-request source. The conflict is the finding — your call which semantics wins.

Half-line: `#toolNames` entries leak when a tool_use_start never reaches input_stop (aborted mid-tool) — bounded by aborts-per-conversation, but a `clear()` on turn end would zero it.

Noted as the agreed follow-up, not a gate: `usage` still lacks `model` (and `costUsd` rides as the attodollar float) — add-only fixes both later, facts-on-the-wire style.

Holding the stamp for rulings on the two findings.

bananabot9000 (approved):
Approved at 13e65e8b, on empirical grounds: the manual-approval test settles it. The wire shows `approval_pending` at 06:33:40.370, then 13.8 seconds of human deliberation, then `approval_settled` at 06:33:54.180 — correct order with real wait-time between. And the auto path in the same capture emits no pending at all (settled as the bare record), which makes the inversion unrepresentable in that mode rather than merely reordered. Both modes project correctly on the current build.

The 4:39 inversion stays on the record as unreproduced-on-current — if it ever reappears, the seam to look at is the two-channel emission (pending via sdkChannel, settled via consumerChannel, no cross-channel sequencing). Turn-granularity finding was withdrawn earlier — the matched pairs with per-request stopReasons are the design working.

Follow-ups already agreed, add-only: `model` on usage; the costUsd float. The conformance kit shipping as real artifacts in the producer's own repo is the tap-testing doc made flesh — the first producer arrives with the bar it has to clear built in.

PR #402, opened 2026-07-05
Stop find | head pipelines from hanging when the consumer exits early

shellicar:
## Summary

- Tear down a pipe stage's upstream when its consumer exits, so `find | head` and similar no longer hang
- Cascade teardown up every hop of a multi-stage pipe, not just one
- Kill a torn-down producer with a real SIGPIPE, so it reports `{ exitCode: null, signal: 'SIGPIPE' }` like any other signal death
- Add a chosen-signal kill to exec-core's process teardown

Supersedes the exploratory #384, whose `runPipeline` was proven in isolation but never run against the package.


bananabot9000 (commented):
Verified at 68355af0, and reviewed with particular care since it supersedes my own exploratory #384 — this is the production version of that design, and it is better than the draft on every axis the draft flagged as rough:

- **Ownership is in the right layer**: the pipe owner (ExecV3's `runPipeline`) does the teardown; exec-core gains only the primitive it should own — a reason-mapped kill. `PipeConsumerGone` living in exec-core keeps the dependency pointing inward. Clean seam.
- **The death is honest**: abort-with-reason → `killSignal` → real SIGPIPE → the producer closes `{ exitCode: null, signal: 'SIGPIPE' }` — the actual broken-pipe death, not my draft's SIGTERM-then-relabel. And the SIGPIPE-handling producer edge is documented with the honest outcome (SIGKILL after grace, reported as such).
- **The draft's open rough edges are all closed**: sink-flush ordering (bridge destroyed *before* the kill so `finished()` can't hang the killed producer's own teardown — the comment walks the exact deadlock), natural exits preserved (`settled[up]` guard), external cancel guarded (`ctx.signal?.aborted` → no synthetic 141s, no double-abort), cascade one hop at a time via settle → teardown → settle.
- **Tested against the real package** — the precise failure of #384, corrected by name: real `Executor`, real `ExecV3` entry, `find | head` promptness, `yes | cat | head` multi-hop with per-stage SIGPIPE assertions, the large-payload flush case, one-result-per-stage.

One finding, and it's exactly the class you told me not to ignore: **a stale comment in the code this PR touches.** `Executor.run`'s pre-abort guard still ends *"Return the same killed status the group-kill path produces (SIGTERM, no exit code)"* — written when the return was hardcoded `'SIGTERM'`, now sitting directly above `killSignal(opts.signal.reason)`, which returns SIGPIPE for a consumer-gone teardown. The parenthetical describes the pre-PR behaviour. One-line fix: "(the reason-mapped signal, no exit code)".

Fix the comment and this is a clean stamp. I'll close #384 with a pointer here — the draft did its job: it proved the mechanism in isolation, and this is what it looks like grown up.

bananabot9000 (approved):
Approved at 2139d287. The parenthetical fixed exactly as flagged — and you swept the whole stale class in the same push (0000272 dropping the leftover 'synthesised 141' wording from the pipe comments, which described the draft's relabelling design rather than the real-SIGPIPE one that shipped). Comments and code cohere everywhere I read. The mechanism verdict stands from the main review: right layer, honest deaths, every draft rough-edge closed, tested against the real package. Clean.

PR #403, opened 2026-07-05
Count tool approval wait as tool time in the status-line clock

shellicar:
## Summary

- Bracket the entire tool-handling method as tool time, so the approval wait counts toward the tools clock
- Fixes the status-line chess clock miscounting time spent waiting for a tool to be approved

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `typescript` as a production dependency so consumers do not need it installed separately
 - Apply biome formatting fixes
+- Count tool approval wait time as tool time in the status-line clock
 - Default `compact.enabled` to `false`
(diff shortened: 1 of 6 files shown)


bananabot9000 (approved):
Approved at 5d641da9, and this is the structural version, not the trigger-move: the bracket IS the method extent. `#handleTools` reduces to toolsStarted → try `#runTools` → finally toolsStopped — so resolution, every approval wait (first and subsequent), execution, all-rejected batches, and thrown errors all land inside tool time by construction. There is no placement left to get wrong: the only caller of `#runTools` is the bracket, and the trigger-sprinkling (`beginToolExecution`, `toolsRunning`) is deleted rather than relocated.

Verified along the way:
- The throw-path leak from my original #394 review (run() throwing past toolsStopped) is fixed as a side effect — `finally` covers every exit.
- The stop-no-op state machine is untouched and the pairs are now clean by construction.
- Comment coherence, finally three-for-three: the old first-run comment is deleted *with* its code, and the new doc states the intent sentence itself — approval waits fall inside the bracket. Code, comment, and what you said from the start now agree for the first time in this feature's life.
- The tests pin exactly the case the old design excluded: the clock opens and closes even when the batch runs no tool.

The residual Unknown between requestSettled and bracket entry is microseconds of dispatch — negligible by design. Clean.

PR #404, opened 2026-07-05
Release 1.0.0-beta.14

shellicar:
## Summary

- Bump all releasing packages to 1.0.0-beta.14
- Catch up the platform package (claude-sdk-cli-darwin-arm64) that beta.13 missed
- Move mcp-exec's wrapped tool from Exec to ExecV3


bananabot9000 (approved):
Approved at 64dc1d53. Release hygiene verified, and the one thing that has bitten a release before — version lockstep — holds:

**All seven releasing packages at exactly 1.0.0-beta.14**: claude-sdk-cli, claude-core, claude-sdk, claude-sdk-tools, exec-core, mcp-exec, and platforms/claude-sdk-cli-darwin-arm64. That last one is the whole point of this PR — it was the package the beta.13 mismatch stranded (the EACCES/pnpm-pack saga), and the launcher's `optionalDependency` on it is `workspace:*`, so both resolving to beta.14 together is exactly what keeps `npm i` from failing on a phantom platform version. The publish-package 'verify version matches tag' step is the enforcement, and this bump satisfies it by construction.

**ci.yml exclusion is a real correctness fix, not cosmetic**: `--filter='!./platforms/*'` keeps the platform packages out of the affected-build matrix — they have no build task and publish only through the SEA path, so without this an affected release that touches a platform pkg would spawn a build job for a package with nothing to build. Matches the #398 design (platforms are wrapped, not built).

**mcp-exec Exec → ExecV3 swap is clean**: symmetrical interface (same defineTool shape — `.description`/`.input_schema`/`.handler`), the `./ExecV3` export exists, MCP validates against `input_schema` before the handler so the `as ExecInput` cast is gated by ExecV3's own schema. Integration test updated to the `commands` shape (from `steps`), including the rm-rf refusal case now asserting on the text channel. Breaking for mcp-exec consumers (schema shape changes) but that's the intent, and it's a beta.

**The CLAUDE.md doc even got corrected in the same PR** — the platform-bump rule now points at the CI `publish-package` verify step rather than the old `publish-platforms` job name. Doc tracks the pipeline it describes.

The honest boundary, same as #398: this PR makes the *versions* correct; the publish:true path itself sits its first real exam when the tag fires. Everything a static review can verify is clean — so if it doesn't bloody work, it'll be the release event, not the diff.

PR #419, opened 2026-07-15
Default the user CLAUDE.md and SYSTEM.md sources off

shellicar:
## Summary

- Default `claudeMd.sources.user` and `systemPrompt.sources.user` to `false`, so the user-level `~/.claude/CLAUDE.md` and `SYSTEM.md` are no longer silently merged into a session at launch
- Makes `--claudeMd` the sole context channel: a composed session gets everything through it, a bare session is genuinely empty
- Project, projectClaude and local sources unchanged; setting user back to `true` in config remains an explicit opt-in


bananabot9000 (approved):
Approved at d0b2d73b. Small and clean — every default/catch pair flipped consistently for both claudeMd.sources.user and systemPrompt.sources.user, project/projectClaude/local untouched, tests updated to match (including the invalid-value fallback and wrong-type cases, which correctly now assert false). Explicit opt-in via config still restores the old behaviour. Nothing else moved. Green.

PR #420, opened 2026-07-15
Add a Skill tool and inject an available-skills catalogue

shellicar:
## Summary

- Add a load-only `Skill` tool that resolves a skill by name from the configured roots and returns its body (frontmatter stripped); an unknown name comes back with the list of what is available.
- Add a `skillDirs` config setting — an ordered, replacement-only list of skill roots, later roots winning on a name collision, empty resolving nothing.
- Inject an available-skills catalogue (name, plus its frontmatter description when present) as a cached `<system-reminder>` on the first user message, scanned once at startup and re-injected after compaction, so the model discovers skills to load without a `list` round-trip.
- Render the tool call as `Skill(<name>)` in the TUI, matching how the other tools show their argument.

The catalogue rides the existing CLAUDE.md `cachedReminders` path, so it sits inside one cache breakpoint rather than adding a fifth. Discovery stays a stable cached prefix and the tool only loads; which skills exist and what is always-on remains launcher policy.

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add per-source CLAUDE.md loading control
 - Add the --system-identity flag: bind a system prompt to a conversation from a file, persisted and restored on resume
 - Add the Memory tool: a persistent, shared, relevance-searchable memory Claude reads and writes across sessions
+- Add the skillDirs config setting: an ordered, replacement-only list of skill roots the Skill tool resolves across, with later roots overriding earlier ones and an empty list resolving nothing
 - Add tools config to select execution tools; ExecV2 enabled by default, Exec (V1) off
 - Add web search and web fetch as built-in server tools
 - Allow --file to be specified multiple times; files attach in argument order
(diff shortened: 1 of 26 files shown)


bananabot9000 (approved):
Reviewed the diff in full (728 lines, ~15 files). Clean, tightly scoped, well tested.

**Design:** load-only Skill tool + separately-injected catalogue is the right split — discovery stays in a stable cached `<system-reminder>` prefix (no `list` round-trip breaking the cache), the tool only resolves name → body. `skillDirs` as replacement-only config (no merged built-in default, empty = visibly bare) matches the same pattern used elsewhere in this codebase (e.g. claudeMd sources).

**Traced the two things I'd actually worry about:**
- Frontmatter fence parsing (`frontmatter.ts`): matches at the *first* column-0 `---`, including one sitting inside a quoted YAML value — deliberately defined behavior per the test comments, not accidental, and consistent with how a real YAML parser would fail on that input anyway (it's not valid frontmatter either way).
- Catalogue building swallows bad YAML per-skill (`readDescription`'s try/catch) so one broken SKILL.md can't drop every other skill from the list — correct fail-open direction for something that gates discoverability.

**Two minor non-blocking observations, not asking for changes:**
- The Skill tool re-scans `skillDirs` from disk on every single invocation rather than caching the resolved map — fine for correctness (guards staleness), just means N invocations = N scans.
- `resolveSkillCatalogue()` is a one-shot startup scan; a `skillDirs` config change mid-session won't be picked up without a restart — already flagged as a deferred concern in your own comment, not something I'm raising as new.

Test coverage is genuinely good — the block-scalar description and embedded-`---` cases are exactly the ones I'd have gone looking for if they weren't already there. Approving.

PR #421, opened 2026-07-15
Bound the primary-view block dividers to a minimum width

shellicar:
## Summary

- Block header dividers now pad to a fixed minimum width instead of the full terminal width, so the trailing run of hyphens no longer scales with the window while short headers still line up.
- Keep the markdown `---` rule deliberately narrower than the block divider: the model's own in-response separator stays visually distinct from the app's block and mode boundaries, so a `---` mid-response never reads as a new block starting.
- Colocate both widths in `model/dividerWidths.ts` with a note on why they differ, so the two constants can't drift or be mistakenly unified.


bananabot9000 (approved):
Small, clean, verified: `target = Math.min(MIN_DIVIDER_WIDTH, cols)` correctly caps at the terminal width on narrow screens instead of overflowing, and the dividerWidths.ts comment explaining why HR_WIDTH and MIN_DIVIDER_WIDTH are deliberately different (model's own `---` vs the app's block boundary) is exactly the kind of thing that stops someone 'helpfully' unifying them later. Tests updated to match. Approved.

PR #422, opened 2026-07-15
Mouse-wheel scrollback for the primary transcript

shellicar:
## Summary

- Scroll the conversation transcript back with the mouse wheel or PageUp/PageDown to read output that has scrolled off screen; only the transcript moves while the editor and status bar stay pinned and live, so you can read a previous response while composing.
- A divider marks how far you are scrolled, and scrolling to the bottom resumes following. Nothing auto-snaps: incoming output and typing both hold your position.
- Wheel events are parsed from raw stdin ahead of readline (which shreds SGR mouse sequences into per-character keypresses), emitting scroll actions and swallowing clicks and drags; a bare Escape passes straight through so cancel is untouched.
- Position holds as new output streams in at the same width, and re-anchors to the same line across a resize by scaling with the rewrap.

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Render assistant responses as styled markdown in the terminal
 - Retry on internal server error
 - Retry transient API errors with exponential backoff and jitter before surfacing the error
+- Scroll the conversation transcript back with the mouse wheel or PageUp/PageDown to read earlier output; the editor and status bar stay pinned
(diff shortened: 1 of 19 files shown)


bananabot9000 (approved):
Read the full diff (985 lines). ScrollState's measure()/offset math checks out — same-width holds absolute position, resize scales by rewrap ratio, clamped to maxOffset either way. extractMouseSequences handles the split-across-chunk-boundary case properly (tested), and the bare-Escape-passes-through guard means cancel still works. One thing worth knowing, not fixing: enabling xterm mouse mode (1000+1006) means the terminal's own native click-drag text selection gets captured by the app instead — universal tradeoff of any TUI that turns on mouse reporting, most terminals let you hold Option/Alt to bypass it for native selection. Not a bug, just worth a one-line mention somewhere users would find it. Approved.

PR #423, opened 2026-07-15
Harden audit-file reads against non-assistant lines

shellicar:
The status line derives its token and cost figures by reading the session's audit file (`AuditStats.derive`, the only reader of `~/.claude/audit`). A separate feature will start writing user-role lines into that file. Those lines carry no `usage`, so the derivation threw a TypeError the moment it met one and the whole status line broke.

Parsed lines now pass through an `isUsableLine` type guard before being summed: an allowlist on role (a missing role is treated as a legacy assistant line, so it still counts) plus a `usage` shape check. The guard narrows to `AuditLine`, which also removes the `JSON.parse(...) as AuditLine` cast that was asserting a shape the data doesn't always have.

Changelogged as a user-visible fix even though this is the app rather than a consumed library.


bananabot9000 (approved):
Small, clean, correctly scoped. isUsableLine's allowlist-with-legacy-default (missing role -> treated as assistant) matches the stated backward-compat need, and the 4 new tests cover exactly the cases that matter: user-role skipped, no-usage skipped, unknown-role skipped, all-skipped-returns-zero. Approved.

PR #424, opened 2026-07-15
Add a SystemReminder type and a live skill-catalogue delta

shellicar:
## Summary

- Introduce a `SystemReminder { text, persisted, position }` type so a per-message reminder states what it is instead of leaving the shape implicit. Persisted reminders are written to history once per query and lead the user message, so they are cached; ephemeral ones (the git delta, the clock stamp) ride the request clone every turn and trail it, staying uncached so the clock never shows on the user message and then vanishes on the model's own turns. `PerQueryInput.reminders` and `TurnInput.ephemeralReminders` replace the old `systemReminder` string.
- Inject a live skill-catalogue delta. `SkillCatalogueTracker` re-scans the skill roots per query, hashes each `SKILL.md`, and prepends a reminder naming the skills that changed. The hash covers the whole file, so a body-only edit still surfaces, since a skill loaded earlier in the session may now be stale. The first scan records the baseline and stays silent, so a resumed session announces nothing.

The CLAUDE.md prefix path and the cache markers are left untouched. The full catalogue still goes where CLAUDE.md goes, and the delta rides the existing moving last-message marker, so the four-breakpoint budget is unchanged.


bananabot9000 (approved):
This is exactly the deferred limitation I flagged on #420 (skillDirs changes needing a restart) — good to see it land. Traced the correctness carefully:

- SkillCatalogueTracker: hash-based (not line-based) diffing correctly catches body-only edits the catalogue line wouldn't show. First-scan-silent is right for both fresh and resumed sessions (the tracker's own `#hashes` starts null regardless of conversation history, so it never fires a false delta on a resume).
- The persisted-leading skill-delta and the cachedReminders (full catalogue) injection are mutually exclusive by construction: cachedReminders only injects on the conversation's very first-ever user message, and the tracker's first scan (which always returns null) always coincides with exactly that query. So they never collide in the same message, and cacheClaudeMdPrefix's cache-boundary marker (which operates on that first message) is unaffected by the new leading/trailing reminder plumbing on later messages.
- SystemReminder's persisted/position axes are clean — ephemeral+trailing (git delta, clock stamp) vs persisted+leading (skill delta) covers the actual use cases without over-generalizing.

One tiny nit, not blocking: SkillCatalogueTracker's `ConfigLoader<any>` loses the config typing DurableConfigFactory otherwise has — probably fine, just noting it. Approved.

PR #425, opened 2026-07-15
Release 1.0.0-beta.16

shellicar:
## Summary

- New Skill tool with an injected, live-updating catalogue of the available skills
- Mouse-wheel scrollback in the transcript, and a running session can move to a new working directory from command mode
- Fixed the TypeScript tools — tsserver now runs on demand instead of going stale
- Injected content wrapped in XML markers; user CLAUDE.md and SYSTEM.md sources now default off
- Cleared out the removed legacy claude-cli app's lingering references


bananabot9000 (approved):
Straightforward release housekeeping — version bump across packages, legacy claude-cli app references cleared from CLAUDE.md/knip.json/launch.json now that it's fully retired, and the old apps/claude-sdk-cli/CLAUDE.md planning doc removed (its content — ref-swapping, Ref store, tool-result-size protection — is long since built, not a live plan anymore). Nothing risky here. Approved.

PR #426, opened 2026-07-16
Tell the model the working directory

shellicar:
## Summary

- New `CwdTracker` emits a persisted-leading `<system-reminder>` naming the working directory, wired alongside the skill-catalogue delta in `runAgent`.
- Announces the current cwd on resume, where the skill tracker stays silent: a stale cwd is a correctness lie, since every relative path the model reasons about resolves against it.
- Announces the from/to when the cwd changes mid-session; a fresh conversation is primed silent until it moves.
- On resume we announce where we are now rather than recover the prior cwd from history, which compaction can evict.

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show working directory name in status bar
 - Support reading PDF and image files as native API content blocks
 - Survive a mid-turn network drop: keep the machine awake during a request, persist the conversation as each message is sent and answered, and resume an interrupted turn from an empty submit
+- Tell the model the working directory: state it up front, and report the from/to when it changes mid-session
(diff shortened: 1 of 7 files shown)


bananabot9000 (approved):
Clean, small, mirrors CwdTracker/SkillCatalogueTracker's now-established pattern well. Correctly justified divergence from the skill tracker: cwd announces even on the FIRST scan (including resume), because a stale cwd is a correctness lie for every relative path the model reasons about, whereas the skill tracker staying silent on first scan is right (the full catalogue is already in context). Traced whether this breaks the cachedReminders/cache-boundary interaction from #424 — it doesn't: cacheClaudeMdPrefix marks its breakpoint at a fixed count of the original cachedReminders blocks regardless of what other persisted-leading blocks get appended after it, so the cwd (and skill) reminders just sit outside that specific marked prefix on the one query where both co-occur. Not a bug, at most a marginally smaller cache window on that one first message. Tests cover first-scan, no-op, and both directions of a move. Approved.

PR #427, opened 2026-07-16
Model selector in command mode

shellicar:
## Summary

- Add a model selector to command mode: `Ctrl+/ m m` opens a free-text model editor pre-filled with the effective model, mirroring the cd path editor.
- Free text always submits and sends — a blue highlight is purely advisory, lighting up only when the typed id exactly matches a known model, and never gates submission.
- The command-mode override shares one slot with the `--model` startup flag; an empty submit clears it back to the config model, whether a human or the flag set the value.
- Add `IModelCatalog`/`ModelCatalog` to the SDK: a lazy, memoised fetch of the account's models from Anthropic's `/v1/models` over the existing OAuth transport, degrading to an empty (advisory-only) list on failure. Built as reusable infrastructure for future token/support-settings work.
- The per-session override is in-memory only, gone on restart, like `--model`.

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `thinking` config (enabled, effort) for extended thinking
 - Add a --claudeMd flag to contribute a string to the assembled CLAUDE.md content at launch
 - Add a history view to navigate and inspect past blocks in the active session
+- Add a model selector to command mode (Ctrl+/ m m): free-text model entry that always sends, with a blue highlight when the typed id matches a known model. Shares one override slot with the --model flag; empty submit clears back to the config model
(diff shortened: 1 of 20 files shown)


bananabot9000 (approved):
Read the full diff. Clean state machine mirroring the cd editor (context, editor buffer, escape/enter handling), and the override-slot logic in ModelOverrides is correctly subtle: `#modelTouched` distinguishes 'never set at runtime' (falls back to --model) from 'cleared at runtime' (falls back to config model) — one slot, two distinct empty states, handled right.

ModelCatalog: the advisory-only contract is actually enforced, not just claimed — submitModel never consults knownModels, so typing and submitting before the catalogue promise resolves (openModelEditor opens the editor synchronously, then awaits the fetch) still always sends. Memoisation is race-safe (cache set before the in-flight promise's finally clears it, so no window for a duplicate fetch), and failure correctly stays uncached so a later call retries — tested directly.

Nothing to flag as a bug. Approved.

PR #428, opened 2026-07-16
Merge PreviewEdit into EditFile: one call, plain-text line-numbered diff

shellicar:
## Summary

- Collapse `PreviewEdit`/`EditFile` into a single `EditFile` tool: validate, apply, write, and return a diff in one call, dropping the in-memory patch store and preview/confirm step
- Return the diff as plain, line-numbered text instead of a JSON-wrapped object, so Claude reads it without unescaping
- Split `replace_text` into a literal string replace (no regex involved) separate from `regex_text`; `textEdits` errors now name the failing edit's index
- Replace the `append` field with negative `after_line` indices (`-1` = end of file), so appending composes with other edits in the same call instead of being mutually exclusive
- Update CLI wiring, docs, config schema, and tests for the merged tool


bananabot9000 (commented):
Traced the write path first since this is the one that matters most: it's atomic — validate → apply (in-memory) → applyTextEdits (in-memory, can throw) → one `fs.writeFile` at the end. Any error throws before the file is ever touched. Good.

One real, verified bug though: appending to an EMPTY file via `after_line: -1` injects a spurious leading blank line. `''.split('\n')` yields `['']` (one phantom element, not zero), so `resolveAfterLine(-1, 1)` resolves to `1`, and `applyEdits`'s splice inserts *after* that phantom line rather than replacing it — result is `"\nnew content"` instead of `"new content"`. I reproduced this by hand-tracing `resolveAfterLine`/`applyEdits` against an empty-file input.

The old dedicated `append` field avoided this entirely (`baseContent + input.append`, no line-splitting). The deleted `EditFile-append.spec.ts` had an exact test for this case ("append to an empty file writes the content") that wasn't replaced with an equivalent `after_line: -1` version — the new suite's closest test ("inserts after the last line when after_line is -1") uses a 2-line file, not empty, so this regression isn't caught.

Everything else checks out clean: negative after_line composes correctly with other edits in the same call (resolved against the original line count, not the mutating result array), replace_text is genuinely literal (no regex path), ambiguous-match handling is explicit and tested, error indices are 0-based and unambiguous, and createEditFilePair.ts is fully and correctly removed (zero remaining references anywhere). Just the empty-file append case to fix.

bananabot9000 (approved):
Verified — traced the fix directly. `const baseLines = baseContent === '' ? [] : baseContent.split('\n');` (with a comment naming the exact bug) means an empty file now resolves after_line against a 0-line total, resolveAfterLine(-1, 0) = 0, and the splice lands at index 0 with no phantom element ahead of it — no leading blank line. New test ('appends to an empty file without a spurious leading blank line') covers exactly this case. Fixed properly, approved.

PR #429, opened 2026-07-17
Split a tool turn into tool-use and execution blocks

shellicar:
## Summary

- Tool turns now render as two blocks instead of one: **tool use** (the model's request) and **execution** (the run). The old single block sealed the moment the model finished asking, so the execution phase and the approval wait fell outside any block's timing; the execution block now covers the real run.
- The SDK emits `tool_exec_start` / `tool_exec_end` around tool execution (bracketing the approval wait and no-op batches), so a consumer can frame and time the run separately from tool-call generation.
- Both the primary and history views project the split: input on the use block, input plus output on the execution block. History navigates the execution block as a tool card.
- Fixes a batch tool-approval bug where a single Y or N settled every tool in the batch at once; a manual answer now settles only the tool it was given for.
- Tests: SDK emit-order bracket, the block split and its history rendering, and an approval regression asserting one answer settles one tool.

The token/cost attribution rework (input on the query block, output on the last block) is intentionally left as separate work, since it touches the status line.

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -85,6 +85,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show model version alongside model name in the status bar
 - Show tool input JSON as it streams
 - Spawn the TypeScript server on demand for each tool block and tear it down after, replacing the always-on server that ran for the whole session
+- Split a tool turn into two transcript blocks, tool use (the model's request) and execution (the run), so the execution block's timing reflects the actual run including the approval wait rather than only the tool-call generation; both the primary and history views show input on the use block and input plus output on the execution block
 - Split model identifier into name and version for separate use
 - The --verify check now boot-checks the tsserver with a one-shot spawn instead of only looking for its path
 - The user-level CLAUDE.md and SYSTEM.md sources now default off, so nothing is silently concatenated into a session at launch; project, projectClaude and local sources are unchanged, and setting user back to true in config remains supported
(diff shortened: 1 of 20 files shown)


bananabot9000 (commented):
Traced the approval-batch fix carefully since it's safety-relevant — it's correct for the bug it targets. Old code unconditionally re-checked `hasPendingApprovals` after every settlement and drained again regardless of which side of the race won; new code only drains when `wonBy === 'wire'`, so a local keypress win no longer double-drains. The new regression test genuinely exercises a 2-tool batch + one local answer and would fail against the old code — real coverage, not a trivial case.

But there's a narrower, untested variant of the same bug class still open: `resolveNextApproval` has no requestId/tool affinity at all — it's a bare `#pendingApprovals.shift()`, pure FIFO (verified in `ToolApprovalState.ts`). So the fix's own comment ("draining here would resolve a different tool's pending approval") is still true on the wire-win path itself, just one layer down: in a multi-tool batch where a WIRE approval settles for a tool that isn't the one at the front of the local FIFO (plausible with the NATS remote-approver feature, where a remote approver could decide tool B before tool A even though A was queued first), draining the head resolves A's local promise with B's answer. Same failure mode, different trigger.

This path is completely untested — every test in the spec file registers `CapturingBus` (no broker, wireAnswer never resolves), so `wonBy === 'wire'` never actually runs in the suite. Worth a requestId-scoped resolver (or matching the drain to `msg.requestId` specifically) before calling the batch-approval class of bug fully closed — not blocking this PR necessarily, since the keypress case (the one actually reported) is genuinely fixed, but flagging it so it doesn't get filed as a surprise later.

bananabot9000 (approved):
Converting to an approval — the block-split work is solid and I've verified the keypress-path approval fix within #429 itself is correct. The gap I flagged (untested wire-path FIFO variant) is a real one, but #430 closes it completely via a different, better mechanism (id-keyed Map, no wonBy tagging needed at all) — so no action needed on #429 itself for that.

One thing worth knowing: #429 and #430 are both branched off main independently, not stacked — #430's diff is based on a pre-#429 version of AgentMessageHandler.ts (no wonBy tagging), so whichever merges second will likely hit a conflict in that function and need a rebase to reconcile #429's wonBy-tagged approach with #430's id-keyed one. Not a blocker, just flagging the merge order will need a manual look.

PR #430, opened 2026-07-17
Fix batch tool approvals settling the wrong tool

shellicar:
## Summary

- Local Y/N approval was anonymous: it settled the head of a FIFO queue, not the tool you'd selected, so in a batch a keypress could answer the wrong tool — or two at once.
- Key the local approval queue by request id (the tool_use id), mirroring the wire path which was already correct; a keypress now settles the selected tool, and each handler drains only its own id.
- Wire/Tower answers were unaffected — this was purely the local keypress path.
- Add an end-to-end regression test driving a real keypress through ApprovalHandler and asserting the tool_approval_response identity; verified it fails on the old behaviour.

Milestone: 1.0


bananabot9000 (approved):
This actually closes both bugs — the reported keypress-selects-wrong-tool issue AND the latent wire-path variant I flagged on #429 (where `resolveNextApproval` was a bare FIFO shift with zero requestId affinity, so a wire answer landing out of local-queue order could settle the wrong tool). Moving `#pendingApprovals` from an array to a `Map<string, resolve>` and keying every resolve call by `requestId` fixes both paths at once, even though the description frames it as purely the keypress fix — the wire drain in AgentMessageHandler.ts is now `resolveApproval(msg.requestId, ...)` unconditionally (not gated on which side of the race won), which is exactly the requestId-scoped resolver I'd have asked for. Called on an already-settled id it's a safe no-op (the entry's already deleted), so the unconditional call doesn't reintroduce anything.

Test coverage is genuinely thorough — order-independence, cross-tool isolation on both the state layer and the full ApprovalHandler-through-AgentMessageHandler wiring, resolveSelected vs queue-head explicitly distinguished. Nothing left open on the approval-identity front that I can see. Approved.

PR #431, opened 2026-07-17
Show token usage on every turn, split into input and output

shellicar:
## Changes

- Token usage shows on every turn, not just tool turns.
- Each turn's usage is split into input and output, priced separately.
- Token totals and context percentage stay correct.

### DIFF
--- a/apps/claude-sdk-cli/src/controller/AgentMessageHandler.ts
+++ b/apps/claude-sdk-cli/src/controller/AgentMessageHandler.ts
@@ -1,4 +1,5 @@
 import { relative } from 'node:path';
+import { RESET } from '@shellicar/claude-core/ansi';
 import { ConfigLoader } from '@shellicar/claude-core/Config/ConfigLoader';
 import { IFileSystem } from '@shellicar/claude-core/fs/interfaces';
 import { ILogger } from '@shellicar/claude-core/logging/ILogger';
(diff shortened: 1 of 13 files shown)


bananabot9000 (approved):
Traced the frame-splitting logic carefully since double-counting or clobbering is the obvious risk with any delta-based accounting change.

Correct: `subtractUsage(total, start)` isolates the output-only delta for the end frame since input/cache are fixed once the request is sent (only output grows during the stream) — so cache/input fields on the end frame should evaluate to exactly 0, which the tests confirm directly ("does not repeat the input tokens on the message_end frame"). `StatusState`'s `if (context > 0)` guard correctly stops the output-only frame from clobbering `#lastContextUsed` to zero — genuinely necessary now that two frames land per turn instead of one, and there's a direct test for exactly that clobber scenario.

Also a real pricing fix riding along, not just a display change: the old code priced a delta (marginal input/cache) and a total (full output) in the same `calculateCostSplit` call, which was conflating two different kinds of quantity. New code prices each frame's cost from only what that frame actually carries — cleaner and more correct.

One tiny thing, not blocking: `#appendUsageLine`'s no-active-block branch hand-indents with a hardcoded 3-space string matching `CONTENT_INDENT` in a different file — a magic-string duplication that could silently drift if that constant changes elsewhere. Approved.

PR #433, opened 2026-07-17
Document PR body and platform-release rules in CLAUDE.md

shellicar:
## Summary
- Release step 6 excludes `platforms/*`: those packages are published by the `claude-sdk-cli` release, not by a standalone tag.


bananabot9000 (approved):
Doc-only, clear and consistent with the actual release process (matches the pattern beta.17's own minimal one-line body followed). The platforms/* exclusion is a real, useful gotcha to have written down — worth documenting exactly because it's non-obvious and fails loudly if done wrong. Approved.

PR #434, opened 2026-07-17
Add conversation-history search to the CLI

shellicar:
## Summary

- Search past conversations by full-text, with citations to the source
- Read a citation back with the surrounding turns
- Write each committed turn to the history index live
- Migrate old audit files to the new format
- Rebuild the index from the audit with a standalone script
- Collapse near-duplicate turns with a background dedup sweep


bananabot9000 (commented):
Big one, reviewed carefully given the scope. Confirmed the file-naming thing myself first: `ingest-history.ts`/`migrate-audit.ts` are thin CLI entrypoints wrapping the camelCase implementation files — deliberate, not leftover duplicates.

**Real gap, verified directly:** `ingestHistory.ts` has no corruption isolation, unlike its sibling `migrateAudit.ts`. `parseAuditLine()` does a bare `JSON.parse(raw)` with no try/catch, and the loop in `ingestHistory()` doesn't wrap the call either — so one truncated/corrupt line in any single audit file throws uncaught and aborts the rebuild for every conversation, not just the bad one. This script's whole purpose is disaster recovery, and `migrateAudit.ts` already has the right pattern next to it (per-session try/catch, tracks `summary.corrupt`, leaves the file untouched) — `ingestHistory.ts` just doesn't use it. Given the tool's most likely to be run right after a crash (exactly the scenario that produces truncated lines), this feels worth fixing before it ships, not just noting.

**Worth knowing, lower severity:** the dedup sweep's near-duplicate threshold has no test for the adversarial case that actually matters for a CLI history — two genuinely different turns that differ by one salient token (a filename, a test name) but are otherwise boilerplate-similar. The "keep unique terms searchable" mitigation only re-indexes the non-shared tokens, so a wrongly-collapsed turn becomes findable only by its unique vocabulary, never by the shared terms someone would naturally search on. Not necessarily wrong, just untested on the case that would actually bite.

**Solid, no action needed:** migration itself (dry-run genuinely defaults on, corrupt lines caught per-session and left untouched, race-checked atomic swap, good test coverage — best in the PR), write-through failure handling (verified the SQLite insert is synchronous so the try/catch genuinely catches a store failure, turn isn't faulted), and SearchHistory→ReadHistory citations (collapse only strips FTS terms, never deletes the rows read() needs, so no dead-citation risk).

Everything except the ingestHistory gap and the dedup test gap looks genuinely solid for something this size.

bananabot9000 (approved):
Verified the fix — parseAuditLine's call is now wrapped in try/catch, corrupt lines are counted and skipped rather than aborting the file, and the new test ('skips a corrupt line and indexes the rest of the file') exercises exactly the scenario I flagged: a truncated line sandwiched between two good ones, both good ones still get indexed. Approved.

PR #435, opened 2026-07-17
Show CLI build version in the status bar

shellicar:
## Summary
- Status bar shows the CLI's own build version, dimmed, at the end of the line


bananabot9000 (approved):
Simple, cosmetic, low risk — build version appended dimmed at the end of the status line in both branches (model set or not), all existing tests updated consistently plus two new ones covering the addition directly. Approved.

PR #437, opened 2026-07-18
Split GitHub access into a read-only path and an approval-gated escalation path

shellicar:
## Summary

- GitHub operations run on two paths: read-only by default, escalated for PR actions.
- Escalated actions always ask for approval first.
- New PRs open as drafts by default.
- PR creation and edits can set milestone, reviewers, assignees, and labels.


bananabot9000 (commented):
Given the incident this closes, I traced the two most safety-critical claims directly rather than trust the intent comments.

**The escalation gate itself is solid.** `operation: 'escalate'` is a hardcoded, model-unreachable field on the six PR tools; `permissions.ts` routes it to unconditional Ask before any auto-approve config can touch it; draft is hardcoded true with no override path; the schema has no `approve` review type at all. Nothing bypasses that gate.

**Two real gaps in the ORDINARY exec path, which is what the reader/holder split depends on:**

1. `buildEnvFrom`'s own comment says cmdEnv is "merged last, so it can override but never supply a resolver's key with a stolen value it doesn't have" — but that reasoning only holds if the model has no other credential to substitute. `cmd.env` is a fully model-settable schema field on every ExecV3 call, and it's spread LAST over the injected reader token (`{ ...env, ...cmdEnv }`) — so any exec call can override GH_TOKEN with whatever the model supplies. Not exploitable today if the model genuinely has no other token available, but it's a live door, not a closed one, and worth knowing before assuming the strip+provide mechanism is airtight.

2. **The bigger one:** the comment at line ~156 says the reader credential is wired "read-only, Contents read-write only, no Pull requests permission." That's not read-only — it's write-scoped for repo contents, just without PR-API access. Meaning: this PR closes the "opened a PR" half of the original incident (now requires the holder token + mandatory approval), but "committed and pushed" is STILL possible under the reader token, gated only by the ordinary exec write-permission matrix — the same kind of governance that let the original incident happen. Worth confirming explicitly whether that's an intentional scoping choice (push is fine, only PR-opening needed the hard gate) or an oversight, since "read-only" as a name implies more than what the token can actually do.

Everything else — draft-by-default, the new keychain-native module (narrow, fixed-argument usage, no obvious widening today), commandMatches/ExecV3 touches being genuine shared-mechanism reuse not scope creep — checked out clean.

bananabot9000 (approved):
Verified both fixes directly.

**Finding 1 (env override) — fully fixed.** `buildEnvFrom` now merges cmdEnv FIRST, strips protected keys from that merged result, then applies `provide` LAST so the forced identity always wins regardless of what the model supplied. The comment even documents that this was confirmed exploitable during the fix (bogus token produced a 401 instead of the real token's 403, proving the caller's value had won) — good that it was actually tested, not just reasoned about.

**Finding 2 (read-only labeling) — clarified as intentional, now honestly documented.** The reader token still holds Contents:read-write (push remains possible via the ordinary exec matrix) — that wasn't changed, and per Stephen it's deliberate: PR-creation is the one action that needed the unbypassable hard gate. The comments now say so plainly ("It is not read-only... the restriction is which operations the credential is authorised for, not a read/write split") instead of the misleading "read-only" framing. Good fix — the design was fine, the description of it wasn't.

Approved.

PR #438, opened 2026-07-18
Give Claude a memory that persists across sessions and projects

shellicar:
## Summary

- Claude keeps what it learns across sessions and projects, from any MCP-speaking client, not only this CLI
- Installing mcp-exec or mcp-memory no longer pulls in the whole Anthropic SDK
- Recording a change and releasing are documented as two separate acts in CLAUDE.md


bananabot9000 (commented):
Checked one specific thing given #440 just had exactly this bug: does the new standalone mcp-memory server's database actually match where the CLI's own memory store lives?

**It doesn't.** The CLI's memory.db lives at `${homedir()}/.claude/memory.db` (via `DatabaseFactory#open()` — confirmed directly). `mcp-memory`'s `openMemoryDatabase()` opens `getDataDir('shellicar-mcp-memory')/memory.db` instead, which resolves to `~/Library/Application Support/shellicar-mcp-memory` on macOS or `~/.local/share/shellicar-mcp-memory` on Linux — a completely different file. So "what one conversation learns, the next can find" wouldn't actually hold between the CLI and this standalone server: they'd each accumulate memories in their own isolated database, never seeing each other's. Same shape as the #440 finding — a new standalone MCP server pointed at its own data dir instead of the CLI's actual store, with nothing bridging them.

Rest of the tool itself looks solid otherwise (didn't need to re-verify what the earlier pass already covered): well-built engine, real WAL/FTS5/versioned migrations, no unintended tool exposure, and the "no longer pulls in the whole SDK" packaging claim checks out against the actual built dist output. The db-path mismatch is the one thing I'd want resolved before calling this "shared" memory rather than "a second, separate" memory.

bananabot9000 (approved):
Verified — the fix is real. `openMemoryDatabase` now opens `${home}/.claude/memory.db` (with an injectable `home` for test isolation), matching the CLI's own DatabaseFactory path exactly, and the comment states the intent plainly: "the same memory.db the CLI itself reads and writes... so a memory written through either one is visible to the other." WAL + busy_timeout carried over correctly for the now-real concurrent-access case. Approved.

PR #439, opened 2026-07-18
Expose TypeScript diagnostics, hover, references, and definitions to any MCP client

shellicar:
## Summary

- Any MCP client can now get type errors, symbol info, references, and go-to-definition without a manual `tsc` run
- Installable on its own, independent of the CLI
- Spawns `tsserver` fresh per tool call and tears it down again after, so nothing lingers as an orphaned process between calls

### DIFF
--- a/packages/mcp-typescript/CHANGELOG.md
+++ b/packages/mcp-typescript/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- TsDiagnostics, TsHover, TsReferences, and TsDefinition tools MCP server, backed by @shellicar/claude-sdk-tools' tsserver bridge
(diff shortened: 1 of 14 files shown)


bananabot9000 (commented):
For the two paths this PR actually controls (success, thrown error), teardown is correctly in try/finally, so the core lifecycle claim holds there.

One genuine open question though, not confirmed either way: `buildTypeScriptService()` builds ONE `tsService` instance shared across the whole process lifetime, and all four tools register against that same instance. If two MCP tool calls overlap (stdio permits pipelined/concurrent `tools/call` requests), and each call's `finally` tears down the shared service once its own promise settles, could one call's teardown land while another call is still using the same instance? The actual reentrancy behavior lives in `TsServerBridge`/`blockEnded()`, which isn't part of this diff (it's existing, unchanged code from claude-sdk-tools) — so I can't confirm this is actually unsafe, just that nothing in this PR's code demonstrates it's safe for the concurrent case, and no test exercises overlapping calls to check. Worth a quick confirmation on whether TsServerBridge already handles concurrent blocks safely, since if it doesn't, exposing this to "any MCP client" (vs the CLI's own single-threaded turn loop) is a new way to actually hit that path.

bananabot9000 (approved):
Verified — this properly closes the concurrency question. Each call now builds its own fresh `ITypeScriptService` inside the handler (no shared instance at all), tracked in an `active` set used as a shutdown backstop for calls interrupted mid-flight. The doc comment states the exact race I was asking about and why the fix removes it. New test ('two overlapping calls both succeed, neither torn down by the other') exercises it directly via `Promise.all`. Timeout raised to 10s with a clear reason (cold spawn is now on the critical path for every call, not just the first). Approved.

PR #440, opened 2026-07-18
Let any MCP client search and read past conversations

shellicar:
## Summary

- Any MCP client can now search and read past conversations, not just the CLI's own model
- Runs as its own local index, separate from the CLI's
- Shares its migration and session-store code with the CLI's own history engine


bananabot9000 (commented):
Traced this one carefully since the PR body claims it's built "from the shared history it produces," and I don't think that's actually true as shipped.

**The MCP server points at a different database than the CLI writes to.** The CLI's own history lives at `${fs.homedir()}/.claude/history.db` (via `DatabaseFactory.getDatabase('history.db')`). This new standalone server opens `getDataDir('shellicar-mcp-history')/history.db` instead — verified directly in `packages/mcp-history/src/entry/index.ts`, a completely different path. Nothing in this PR ingests the CLI's audit logs (or its history.db) into that new location — no ingest script, no reuse of `ingestHistory.ts` pointed at the new path. As shipped, SearchHistory/ReadHistory over this server would return empty results indefinitely, since nothing ever writes to the file it reads from. Worth confirming whether ingest wiring is a follow-up, because right now the feature doesn't look reachable.

**The doc comment and the code disagree on which path is used, too** — the comment literally says "the store directory comes from `getDataDir('claude-cli')`... not a CLI-specific path," but the code two lines down calls `getDataDir('shellicar-mcp-history')`. Different string entirely. Flagging the conflict, not picking which one's the intended behavior.

The good news: the shared migration/session-store code genuinely IS shared, not duplicated — `SqliteHistoryEngine.ts` is a straight rename from the CLI app into `packages/claude-core`, so the corruption-isolation fix from #434 travels with it cleanly, no regression there. And WAL + busy_timeout are already set, so concurrent access isn't a real risk given there's no shared file to contend over yet anyway.

Left as a comment — the empty-database gap feels like it needs a decision (add ingest wiring, or point at the same file as the CLI) before this does what the PR description says it does.

bananabot9000 (approved):
Verified — the doc comment now says plainly what the code does: "this package's own history store... a separate index, not a view onto the CLI's. Nothing currently ingests the CLI's audit logs into it, so the store starts empty until something writes to it." No more disagreement between the comment and the code, and the empty-start behavior is now honestly documented rather than implied to be something it isn't. Approved.

PR #442, opened 2026-07-18
Recover from a conversation left on a dangling tool_use

shellicar:
## Summary
- A conversation that died mid tool-round (crash, kill signal, hung tool) now self-heals on load: any tool_use left without a result gets an honest synthetic error result instead of leaving the API request invalid.
- An auto-denied tool call now tells Claude it was denied by permission policy, not by a user who saw and rejected it.

### DIFF
--- a/apps/claude-sdk-cli/src/controller/AgentMessageHandler.ts
+++ b/apps/claude-sdk-cli/src/controller/AgentMessageHandler.ts
@@ -480,12 +480,17 @@ export class AgentMessageHandler {
         return;
       }
       let approved: boolean;
+      let autoDenyReason: string | undefined;
       if (perm === PermissionAction.Approve) {
         this.logger.info('Auto approving', { name: msg.name });
         approved = true;
       } else if (perm === PermissionAction.Deny) {
         this.logger.info('Auto denying', { name: msg.name });
         approved = false;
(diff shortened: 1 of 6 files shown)


bananabot9000 (approved):
Clean fix, traced both pieces.

healDanglingToolUse() correctly checks the tip is an assistant message with tool_use content before healing (so it's a no-op both on an empty/user tip AND when a real tool_result already answered it, since that answer's row would itself be the tip with role='user', not 'assistant' — no double-heal risk). Handles multi-tool batches by healing every dangling id in one synthetic row, not just the first. The synthetic result is honest about what it claims ("the outcome is unknown", is_error:true) rather than fabricating a plausible-sounding result — matches the discipline this session/history stuff generally holds to.

The auto-deny reason fix is the smaller but genuinely important one: distinguishing "denied by policy, no human ever saw this" from "a human explicitly rejected it" matters because the two should produce different model behavior — a real human rejection is a hard no-retry signal, but a policy auto-deny with the wrong framing could make the model read it as the same kind of finality when it might just need a different tool or approach.

Test coverage is thorough on both — empty/no-tool_use/already-answered/multi-batch cases for the heal, and a direct reason-string assertion for the auto-deny distinction. Approved.

PR #443, opened 2026-07-18
Publish keychain-native as an optional macOS arm64 dependency

shellicar:
## Summary

- `npm install @shellicar/claude-sdk-cli` no longer fails with a 404 on `@shellicar/keychain-native`
- keychain-native is now a real published package, installed only on macOS arm64
- Adds a dedicated native-build CI lane for it, separate from the generic Linux publish path


bananabot9000 (commented):
Found something real, and it's not in this diff at all — it's in what this diff now makes REACHABLE.

**Making keychain-native genuinely optional exposes a runtime break on every non-macOS-arm64 platform.** `EnvProvider.buildEnv()` (from #437, unmodified here) is registered unconditionally in `container.ts` (`services.register(IEnvProvider).to(EnvProvider)` — no platform check) and runs for EVERY ordinary exec call, not just gh-related ones. It calls `this.secrets.ghReaderToken()` unconditionally via `provide: { GH_TOKEN: () => ... }`, and `buildEnvFrom`'s `provide` resolvers run fresh on every single call. `ghReaderToken()` → `readKeychain()` throws when `resolveKeychainNativePath()` returns null — which it always will on Linux/Windows/Intel-Mac, since keychain-native is now correctly os:darwin/cpu:arm64-restricted and won't even be installed there.

So: before this PR, `npm install` 404'd for literally everyone, on every platform (no one could install the CLI at all, so this bug was masked). After this PR, install succeeds everywhere, but running Exec — the core tool, used for everything, not just gh operations — would immediately throw on any platform other than macOS arm64. This isn't a new bug this PR introduces directly; it's a pre-existing gap in #437's EnvProvider that this PR's fix is the first thing to actually make reachable, since before now nobody could get far enough to hit it.

Given Exec is core functionality, this feels like it needs a platform check or a fallback before merge — either EnvProvider degrades to a passthrough (no gh scoping) on unsupported platforms with a clear log message, or the reader-token resolution becomes lazy/conditional (only actually calls into keychain-native if the command looks like it touches gh, not on every call). Happy to be wrong here if there's a guard I'm not seeing, but I traced the actual call chain and don't see one.

bananabot9000 (approved):
Verified — this properly fixes it. `ghScoping` is now opt-in (default false), and even when enabled it's gated by `KEYCHAIN_PLATFORM_SUPPORTED` (darwin+arm64, computed once). When scoping isn't active, `provide: {}` means `ghReaderToken()`/`readKeychain()` is never called at all — so it can't throw on unsupported platforms, full stop. Stripping ambient gh/ssh credentials is correctly kept as an independent, separately-toggleable default-on behavior, so the original threat model (exec shouldn't inherit ambient creds) doesn't regress just because scoping got gated. Config schema descriptions are clear about what happens in each disabled/unsupported/not-set-up state.

One small gap: no dedicated EnvProvider test for the platform-gating branch itself (cli-config.spec.ts covers the schema defaults, not this logic) — but the fix is simple enough (a boolean AND) that I'm confident in it by inspection. Approved, ignoring CI per your note.

PR #445, opened 2026-07-18
Conform to Tower's conversation v2 and agent specs

shellicar:
## Summary

- Moves conversation traffic from Tower's v1 wire shape to v2, including the new query-closed event
- Adds the agent concern: this CLI now announces itself, reports liveness, and answers drain/chdir requests
- Adds conformance schemas and fixtures for both, per Tower's spec


bananabot9000 (commented):
Given Tower's whole philosophy is 'the spec is the only shared surface,' I actually pulled conversation-spec.md and agent-spec.md from the tower repo directly and checked this against the real external contract, not just internal consistency.

**The core protocol implementation is genuinely spec-compliant, not just self-consistent.** Zod schemas are copied verbatim from the spec's normative blocks — no field drift. Subject routing correctly follows v2's leaf-based rule (subject carries the type, not the body). The `say` precondition fix now requires `tip` explicitly (including `null`) rather than treating a missing precondition as always-valid — matches the spec's "the premise is required, not optional" rule precisely. Query-closed fires only after `convChanges.flush()` — after the closing fact is actually committed — matching the committal-ordering rule exactly, and it's genuinely exercised end to end by `producer.conformance.spec.ts` against the real `ConvChangePublisher`/telemetry code, not just schema-validated fixture text. v1→v2 is a clean cutover (checked directly, no live code still publishes `conv.v1.*`); `approval.v1.*` correctly stays untouched since it's a different concern.

**Real gap: the new agent concern (`AgentServicer.ts`, `AgentPresence.ts`) has zero behavioral test coverage.** Verified directly — `servicer.conformance.spec.ts`'s agent-fixture tests only validate the hand-written `scenario-a*.jsonl` TEXT against the JSON Schema; they never drive the real `AgentServicer`/`AgentPresence` code to produce those events and check the actual output. Compare to the conv concern, which has `producer.conformance.spec.ts` running the real producer end-to-end. A regression in the already_attached/unsupported/not_found branching in `AgentServicer` would go completely undetected today — the fixtures prove the SHAPE is right, not that the CODE produces it.

Given how solid the actual wire-conformance work is, this feels like a 'finish the test for the newer half' note rather than a blocker, but wanted it on record given it's the one part of this PR the fixtures don't actually prove.

bananabot9000 (approved):
Verified the test-coverage fix — servicer.conformance.spec.ts now actually drives the real AgentServicer through DI (already_attached/unsupported/drain-accept-and-event/chdir-accept-and-real-move/not_found/unknown-request), not just schema-validating fixture text. That gap is closed properly.

Also traced the detach fix: `cleanup()` now calls `agentPresence.detach(session.id)` before the connection drops, wired into the single shared shutdown path that ctrl+c (via QuitHandler → ShutdownCoordinator), SIGINT, SIGTERM, and the drain event all route through — good, matches the comment's own reasoning ('a clean exit reads as detached, never as silence, which folds to stranded').

One thing I couldn't confirm either way: I don't see an explicit Ctrl-D/EOF handling path anywhere in main.ts or the controller layer in this diff — no `stdin.on('end', ...)`, nothing binding a distinct ctrl+d key. If Ctrl-D exits through some OTHER path that doesn't go through this same `cleanup()`, it'd still miss the detach event even after this fix. Can't verify this one without actually running the TUI interactively — worth a quick manual check if it's not obviously the same path as ctrl+c.

PR #446, opened 2026-07-18
Config startup display shows only the keys it named

shellicar:
## Summary

- The `--config` startup display now shows only the keys the payload actually named, instead of the full merged config
- A key the schema doesn't recognise is dropped rather than shown, since it had no effect
- A nested override (e.g. one hook field) surfaces only that field, not the whole parent object

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -73,6 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Small, clean, well tested. pickOverriddenConfig correctly handles the edge cases that matter: unrecognised key dropped (schema mismatch), unnamed key dropped (not in the override at all), nested partial override recursing to just the named sub-field rather than the whole parent object. Array values correctly skip the recursion path (isPlainObject excludes arrays) and fall through to the resolved value directly, which is the right call for something that's typically replaced wholesale rather than deep-merged. Approved.

PR #447, opened 2026-07-18
Release 1.0.0-beta.20

shellicar:
Bumps two packages to 1.0.0-beta.20.


bananabot9000 (approved):
Clean, minimal — just the two packages with actual changes since beta.19 (claude-sdk-cli from #445/#446, its darwin-arm64 platform package following in lockstep per the rule), plus catching up the CLAUDE.md package table with keychain-native's entry. Nothing in the diff itself to worry about — if 'will this work' is about the actual publish/CI lanes succeeding rather than this PR's correctness, that's not something I can verify from the diff alone, but the version math and package selection here are sound. Approved.

PR #448, opened 2026-07-19
Fix GitVersion silently failing to resolve in CI

shellicar:
## Summary

- Fix GitVersion silently resolving to an empty version in CI: it was extracted into `sudo`-owned `/usr/local/bin`, which the runner user then couldn't execute
- Add a runnability check right after install, so a broken install fails loud instead of silently
- Manual dispatch now runs the whole pipeline (checks/build/wrap), not just the publish jobs, and diffs affected packages from the merge-base with `main`

### DIFF
--- a/.github/actions/install-gitversion/action.yml
+++ b/.github/actions/install-gitversion/action.yml
@@ -17,7 +17,12 @@ runs:
   using: composite
   steps:
     - shell: bash
+      # Installed under $HOME, owned by the runner user, not root via sudo:
+      # a sudo-owned /usr/local/bin binary failed to memory-map when the
+      # runner user later tried to execute it (error 13, EACCES).
       run: |
+        mkdir -p "$HOME/.local/bin"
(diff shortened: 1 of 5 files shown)


bananabot9000 (approved):
Traced the two riskiest parts.

**Root cause and fix are sound:** moving off sudo entirely (installing to $HOME/.local/bin, runner-owned) removes the whole class of permission mismatch rather than patching around it, and the runnability check right after install (in both build-package.yml and checks.yml) means a broken install now fails loud immediately instead of silently producing an empty version three steps later.

**Checked the safety claim on workflow_dispatch, not just trusted the comment:** verified directly in publish-package/action.yml that the dry-run gate on `github.event_name == 'workflow_dispatch'` already existed before this PR — this PR just widens which jobs (checks/detect/build/wrap/verify-keychain-native) also run on dispatch so a manual trigger actually exercises the full pipeline, relying on that pre-existing (and confirmed correct) safety net rather than introducing new publish risk.

The merge-base logic in `detect` for the workflow_dispatch case looks right too — falls back to HEAD^ when dispatched directly on main (merge-base equals HEAD, nothing to diverge from), otherwise diffs from the actual fork point so commits main gained after divergence don't leak into the affected-packages list. Approved.

PR #449, opened 2026-07-19
GitHub_PullRequest_AutoMerge requires a merge strategy

shellicar:
## Summary
- Enabling auto-merge now requires picking a merge strategy (merge, squash, or rebase), passed as its own flag alongside `--auto`
- Disabling auto-merge is unchanged (`--disable-auto` only)


bananabot9000 (approved):
Small, clean, correctly enforced — the .refine() makes strategy required exactly when enable is true (Zod validates before buildArgs ever runs, so the runtime guarantee holds even though the type itself doesn't narrow strategy to non-optional inside buildArgs). buildArgs correctly branches: disable path unchanged, enable path now emits --auto plus the actual --merge/--squash/--rebase flag instead of silently deferring to whatever the repo's default happened to be. Approved.

PR #450, opened 2026-07-19
Allow command mode to be entered, exited, and submitted from any turn phase

shellicar:
## Summary
- Command mode can now be opened, navigated, and closed while a query is streaming, not just while awaiting input
- Submitting with ctrl+enter now works while command mode is open, instead of requiring it closed first
- Attachments added mid-query, and command mode's open/closed state, no longer get reset once that query finishes


bananabot9000 (approved):
Traced the chain wiring and the ctrl+enter passthrough. CommandKeyHandler now sits in both chains, and it explicitly falls through on ctrl+enter (checked before the cdEdit sub-mode branch, so it's unconditional across every command-mode sub-context, not just root) — reaches EditorHandler in the editor chain (submit works), reaches CancelHandler in the streaming chain (correctly a no-op there, since there's nothing to submit mid-stream anyway). Test coverage matches: passthrough at root, passthrough inside the cd editor specifically, and an explicit check that ctrl+enter doesn't also close command mode as a side effect.

The reset() removal is clean too — it was the one command-mode/attachment cleanup call in runAgent's post-query teardown, and it's gone along with its own tests (not left dangling), while every other piece of teardown (tools, editor, phase) is untouched. Matches the stated intent (attachments/command-mode state now genuinely persist across a query boundary) rather than being an accidental side effect of the refactor. Approved.

PR #451, opened 2026-07-19
Show how to resume this conversation when you quit

shellicar:
## Summary

- Quitting the CLI now shows the command to resume this same conversation next time


bananabot9000 (approved):
Small, nice quality-of-life addition. Checked that this doesn't leak into a non-interactive context — --verify runs its own separate early-exit path before cleanup() is even defined, so it's unaffected, and this is the interactive TUI's stdout (not one of the stdio MCP servers where stdout is reserved for JSON-RPC framing), so no protocol-pollution risk either. Approved.

PR #452, opened 2026-07-19
Add duration to ExecV3 results

shellicar:
## Summary
- ExecV3 results now carry a `durationMs`, per command and for the whole run
- A pipe's overlapping stages are measured correctly, so the total isn't just a sum of the parts


bananabot9000 (approved):
Clean, well tested. Clock is properly injected (defaults to performance.now(), overridable for tests), and the deterministic clock-mock test nails the exact arithmetic across the 4-call sequence (evaluate start, stage start, stage end, evaluate end). The overlap behavior is both correctly documented (schema comment explicitly says top-level durationMs is NOT the sum of per-command values) and directly tested with a real slow-producer/fast-consumer pipe proving the total reflects the shorter wall-clock, not the sum. Approved.

PR #453, opened 2026-07-19
Abstract createWriteStream and platform/arch behind IFileSystem

shellicar:
## Summary
- Exec, ExecV2, and ExecV3 redirect writes go through `IFileSystem.createWriteStream` instead of importing `node:fs` directly
- Keychain platform/arch support check reads through `IFileSystem` instead of `process.platform`/`process.arch`
- Adds test coverage for `NodeFileSystem` (previously untested) and fixes `EnvProvider` tests to be deterministic across host platforms

### DIFF
--- a/apps/claude-sdk-cli/CHANGELOG.md
+++ b/apps/claude-sdk-cli/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Hook input delivered via stdin instead of command arguments
 - Internal: split AppLayout into TerminalRenderer, TerminalInput, View, and PrimaryView for future peer views
 - Introduce core-di-lite for dependency resolution; separate composition from logic
+- Keychain platform/arch support check now reads through IFileSystem instead of process.platform/process.arch directly
 - List --file in --help output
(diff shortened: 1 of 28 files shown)


bananabot9000 (approved):
Traced the mechanical parts carefully since a parameter-order change (resolveSinks went from (cmd, stdoutDest?) to (cmd, fs, stdoutDest?)) is exactly the kind of thing that silently breaks a call site someone missed updating. Checked every call site — execCommand, both executeTree call sites, and the 3-arg pump() call (resolveSinks(pipeline, ctx.fs, fwd)) — all correctly match the new signature, nothing passing fwd where fs should go or vice versa.

Genuinely nice test-determinism win too: EnvProvider's ghScoping tests previously computed their own expected value from the real host platform (comment admitted CI runs on ubuntu, dev machines vary) — now they inject a MemoryFileSystem with explicit platform/arch and assert concrete outcomes, covering the actual matrix (supported platform+arch, wrong platform, wrong arch, scoping disabled) instead of one test whose expectation shifted depending on where it ran. NodeFileSystem's new spec file exercises real disk writes (fresh/overwrite/append) rather than just mocking, closing a real gap. Approved.

PR #454, opened 2026-07-19
Add Azure CLI and Azure DevOps privilege escalation tooling

shellicar:
## Summary

- Azure resource management and Azure DevOps PR operations run under certificate-authenticated identities, never a standing personal login.
- Named `AzureDevOps_PullRequest_*` tools mirror the existing GitHub PR tools; `AzCli`/`EscalatedAzCli` give scoped free-text `az` access.
- PR tools resolve org/project/repository from the target repo's git remote and accept an optional working directory.
- Setup scripts provision the underlying service principals, certificates, and permissions.


bananabot9000 (commented):
Given this mirrors #437, checked specifically whether the two lessons from that review carried forward.

**Both did, and #1 is actually stronger than the fix, not just equivalent.** `createAzInputSchema` has no `env` field in the schema at all — the model literally has no surface to override anything, unlike ExecV3 where the fix was ordering-based (inject last). Structurally impossible here rather than correctly sequenced. Escalation gate is also solid: `operation: 'escalate'` hardcoded on every Azure DevOps PR tool and EscalatedAzCli, no `approve` value in the vote schema's enum, same structural (not config-default) gate as GitHub's.

**One real gap, same shape as the GitHub mislabeling, just not yet proven to bite: the reader identity's actual Azure RBAC role isn't enforced or verified anywhere.** `az-sp-create.sh` takes `--role` as a required, free-form operator argument for BOTH the reader and holder accounts — verified directly, no default, no validation. The holder identity gets a dedicated `az-holder-remove-delete.sh` that provably narrows it to a specific safe role after creation (an inspectable, code-enforced artifact). Nothing analogous exists for the reader — "unprivileged" is pure operator discipline at creation time (remembering to type `--role Reader`), not anything the code checks or narrows afterward. If an operator ever fat-fingers `--role Contributor` while provisioning the reader account, nothing here would catch it. Worth a script (or at minimum a hardcoded safe default) before this runs against a real subscription, the same way the GitHub reader token's actual scope eventually got pinned down and documented rather than assumed.

Credential handling and the ADO-remote parsing both check out clean — scripts are dry-run-by-default with the same stdin-secret discipline as the GitHub ones, and `parseAdoRemote.ts` fails closed (returns null) on anything not matching a known remote shape rather than guessing, with parsed values going into az as array args, never through a shell.

bananabot9000 (approved):
Verified — this fully closes the gap, and better than I'd have asked for. `--role` free text is gone entirely, replaced with `--identity reader|holder` mapping to a fixed role (Reader / the custom 'Contributor (No Delete)' role) — there's deliberately nothing else to type, so a fat-fingered role is structurally impossible now, not just validated against. And the holder's safe role is created atomically in this same script at provisioning time (a real custom role definition, not plain Contributor patched afterward) — the comment nails exactly why that matters: a follow-up narrowing script is itself a gap, since it depends on someone remembering to run it. Approved.

PR #455, opened 2026-07-19
Fix wire input deadlock, remove dead code, tighten credential permissions

shellicar:
## Summary

- Fixes a bug where an I/O failure mid-turn could permanently lock remote (wire) input out of a running session.
- Removes dead code that logged raw keystrokes to disk.
- Closes a local information-disclosure gap by restricting stored OAuth credentials to owner-only access.
- Keeps history and memory search behaving consistently by removing duplicated sanitization logic.
- Corrects stale documentation that no longer matches how sessions are saved.
- Fixes a broken internal diagnostic script and guards it with type-check so it can't silently drift again.


bananabot9000 (approved):
Traced the busiest one carefully — the try/catch/finally around runTurn. Confirmed the deadlock scenario is real in the old code (nothing guarded convServicer.setBusy(false), so any throw between session.saveSession() and the end left the servicer permanently busy) and the fix is correctly scoped: turnInProgress/currentAbortController/setBusy(false) all move into finally, guaranteed regardless of outcome. Also checked whether this could double up with or leave a gap against runAgent's OWN error handling — it already has its own try/catch/finally that resets primaryViewState back to 'editor' on any queryRunner.run() failure, so this new outer catch is specifically covering the OTHER stuff in the turn lifecycle (config/git/session I/O), not overlapping or leaving a hole.

Also verified the toFtsMatch dedup doesn't silently change memory search behavior — the surviving implementation (moved from history/search.ts) is byte-identical to the deleted memory/search.ts version, so no regression from consolidating.

The credentials chmod fix is correct and necessary — writeFile's mode option really does only apply on file creation, not on an existing file, so the explicit chmod(0o600) is the only way to lock down a credentials file that pre-dates this fix.

One minor thing, not a bug: tool-schema-sizes.ts's new `const stubFs = null as unknown as IFileSystem` is a bit fragile — safe today since createAppTools never touches fs during construction (only stores it for later handler calls this script never makes), but it'd crash this diagnostic script if that ever changed. Low severity, dev-tooling only, not blocking. Everything else here is genuinely clean. Approved.

PR #456, opened 2026-07-19
Release 1.0.0-beta.21

shellicar:
Bumps six packages to 1.0.0-beta.21.


bananabot9000 (commented):
Caught something given the new bundling rule this PR itself adds to CLAUDE.md ('a bundled dependency... must be bumped and rebuilt... check this by hand, it is not caught by has this package's own src/ changed').

Checked which packages actually bundle their first-party deps (no claude-sdk-tools/claude-core in their own package.json `dependencies` — meaning inlined at build time): mcp-exec, mcp-memory, mcp-typescript, and mcp-history all do. mcp-exec and mcp-memory are correctly bumped here to beta.21. **mcp-typescript and mcp-history are not — both still show beta.18 in this diff.**

Concretely for mcp-history: it imports `SqliteHistoryEngine` from `@shellicar/claude-core/history`, and that exact file's import changed in #455 (`toFtsMatch` moved from `./search` to `../search`) — a real source change to something mcp-history's own dist bundles. Its published dist is stale relative to that change until it's rebuilt and republished at beta.21, exactly the gap this PR's own new rule is meant to catch. mcp-typescript's bundled surface (the TS-diagnostic tools specifically) may not have actually changed tonight, so it's a lower-confidence flag — but given the rule says 'check by hand,' I'd rather flag both and let you confirm than assume tree-shaking makes one of them moot.

Not blocking — the version math for what IS in this diff is otherwise correct — but feels worth catching before merge given it's the exact class of miss the new CLAUDE.md paragraph was just written to prevent.

bananabot9000 (approved):
Retracting my earlier comment — checked whether the change I flagged (toFtsMatch's file move) actually altered behavior, and it doesn't; pure internal reorg, nothing for a rebuild to change for consumers. Version math for what's actually in this diff is correct. Approved.

